PR #2, opened 2018-06-07
Make explicit that this is the MIT license

ericbodden:


pdschubert:
Already added the license explicitly.

PR #19, opened 2018-12-03
add slash after target dir

ryanphuang:
The LLVM installation script assumes target_dir to be specified with a trailing `/`. Not sure if this is intended or not. If I run the script with `./install-llvm-5.0.1.sh 40 ~/software/llvm`, it will clone the projects into `~/software/llvmllvm-5.0.1`. This patch always appends a `/` in the target_dir path, which doesn't hurt even if the target_dir is specified with a `/`.


pdschubert:
It was unintentional - Thank you very much for the fix.

PR #20, opened 2018-12-17
re-enable osx testing

stephengroat:


pdschubert:
Sorry, but we currently have some problems to get the build pipeline running on OSX using travis. We will fix this when we upgrade to LLVM-8.0.0 at the end of february. The bug is partly caused due to different interpretations of the C++ standard (GCC/Clang).

PR #61, opened 2019-09-23
Refactor build for boost, and llvm target dir and running tests.

ddiepo-pjr:
These changes allow the user to specify an existing boost install, and
if not specified now boost installs to a version-specific directory,
keeping it separate from potential conflicts with other versions of
boost installed.  Also, the cmake boost dependencies are always
initiated from a find_package call to ensure we get the desired
boost version.

Similarly, llvm now installs to a version-specific directory as well.

Both these change allow a developer to work on projects which may
require different versions of these dependencies.

Lastly, unit tests can be run as part of the bootstrap, albeit somewhat
crudely. The test PhasarLLVM/Mono/InterMonoTaintAnalysisTest is failing
within TaintTest_03.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,7 +262,7 @@ target_link_libraries(phasar
   ${CLANG_LIBRARIES}
   ${llvm_libs}
   curl
-  gtest
+#  gtest
 )
 
 target_link_libraries(myphasartool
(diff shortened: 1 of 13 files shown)


ddiepo-pjr:
I create a clean docker ubuntu 18.04 image and ran a test compile with these proposed changes.  Good news, everything built without issue _and_ all the unit tests passed.  So my comment above about InterMonoTaintAnalysisTest must have something to do with my local setup.

ddiepo:
@pdschubert , I'm just checking in to see if you've had a chance to look at my suggested changes here and have any feedback.

pdschubert:
As discussed in our email correspondence I am closing this pull request.

Issue #79, opened 2020-01-29
Finish printMethod() to printFunction() Transformation


pdschubert:
To finish our minor refactoring from printMethod() to printFunction() please replace all occurrences of the template typename `M` (for method) by `F` (for function) wherever appropriate. (Use grep to get an idea where `M` is being used.)

PR #86, opened 2020-02-05
printMethod() to printFunction() Transformation

bsharma021:
Changed all template names and used variable names for "methods".
#79


pdschubert:
Thanks!

PR #101, opened 2020-03-11
Removes const version of bits setter function in BitVectorSet

vulder:
Const qualifying a setter functions does not work and leads to compile
errors, therefore, we remove the const version of the setBits function.


pdschubert:
Thanks!

I need to be more careful when merge code ;-)

PR #106, opened 2020-03-25
docker: image build fix

adrianherrera:
Docker image does not build because utils/install-llvm.sh expects 4 arguemnts. The install directory was missing, so I added a sensible default.


pdschubert:
Great, thanks!

PR #107, opened 2020-03-25
docker: entrypoint fix

adrianherrera:
Sorry, another Docker fix that I missed! The entry point directory was incorrect


pdschubert:
Many thanks.

PR #111, opened 2020-03-29
docker: performance improvements

adrianherrera:
Sorry for spamming you with Docker-related PRs!! This will be the last one (I think!).

This PR improves image-build performance. Don't copy everything into the image in one go: this means that any change to any file resets the whole image build (including rebuilding LLVM, etc.). Instead, copy only what is required for the current build step.


pdschubert:
Hey, no worries. We are always happy to receive improvements.

PR #112, opened 2020-03-31
Expose emit JSON

adrianherrera:
Hey Phasar folks!

Here is a non-Docker-related PR 🙂

I noticed that exporting to JSON is not exposed on the CLI. So I added this feature. I find this useful for post-processing Phasar's analysis results.

Thanks for the great tool!


pdschubert:
Awesome! :+1:

PR #117, opened 2020-04-07
Add Arch support to bootstrap

J03D03:
The bootstrap script determines whether pacman is available or not and installs dependencies.
Since there is no 'BuildEAR (bear)' package available via pacman an installBuildEAR script was added to build it from source.


MMory:
Thank you very much.

PR #126, opened 2020-04-21
Revert "Revert "Fix build against LLVM-10 fat shared lib.""

pdschubert:
Reverts secure-software-engineering/phasar#125


pdschubert:
Okay, now we are good to go.

PR #129, opened 2020-04-27
Ninja installation fix

J03D03:
# Changes:
- Update the bootstrap and llvm install script
- Create `/tmp` files only if necessary
- Make use of the script for apt packages 
- Use pip3 packages
- Use `cmake --build` as suggested by @vulder 

## Test with Docker:
Dockerfile :heavy_check_mark: 
```
FROM ubuntu:latest
RUN apt -y update && apt install bash sudo git -y
RUN mkdir -p /usr/src/phasar
WORKDIR /usr/src/phasar
COPY utils/InitializeEnvironment.sh /usr/src/phasar/utils/InitializeEnvironment.sh
RUN /usr/src/phasar/utils/InitializeEnvironment.sh
COPY . /usr/src/phasar
RUN /usr/src/phasar/bootstrap.sh
```

### DIFF
--- a/utils/install-llvm.sh
+++ b/utils/install-llvm.sh
@@ -8,15 +8,13 @@ re_number="^[0-9]+$"
 re_llvm_release="^llvmorg-[0-9]+\.[0-9]+\.[0-9]+$"
 
 num_cores=${1}
-build_dir="${2}"
-dest_dir="${3}"
-llvm_release="${4}"
+dest_dir="${2}"
+llvm_release="${3}"
 
-if [ "$#" -ne 4 ] || ! [[ "$num_cores" =~ ${re_number} ]] || ! [ -d "${build_dir}" ] || ! [[ "${llvm_release}" =~ ${re_llvm_release} ]]; then
-	echo "usage: <prog> <# cores> <build dir> <install dir> <LLVM release (e.g. 'llvmorg-9.0.0')>" >&2
+if [ "$#" -ne 3 ] || ! [[ "$num_cores" =~ ${re_number} ]] || ! [[ "${llvm_release}" =~ ${re_llvm_release} ]]; then
+    echo "usage: <prog> <# cores> <install dir> <LLVM release (e.g. 'llvmorg-9.0.0')>" >&2
 	exit 1
 fi
-
 
 if [ -x ${dest_dir}/bin/llvm-config ]; then
    version=`${dest_dir}/bin/llvm-config --version`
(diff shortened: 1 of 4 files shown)


utils/install-llvm.sh
```diff
@@ -8,38 +8,38 @@ re_number="^[0-9]+$"
...
-
 if [ -x ${dest_dir}/bin/llvm-config ]; then
    version=`${dest_dir}/bin/llvm-config --version`
    echo "Found LLVM ${version} already installed at ${dest_dir}."
    exit 0;
 fi
 
 echo "Getting the LLVM source code..."
-if [ ! -d "${build_dir}/llvm-project" ]; then
+tmp_dir=/tmp/${llvm_release}
```
vulder, line 26:
```suggestion
tmp_dir=$(mktemp -d -t phasar-llvm-${llvm_release}-XXXXXXXXXX)
```
This makes it a bit more robust in cases where the system defines a different `tmp` or in case of clean-up failures.

utils/install-llvm.sh
```diff
@@ -8,38 +8,38 @@ re_number="^[0-9]+$"
...
 fi
+
 echo "Building LLVM..."
-cd ${build_dir}/llvm-project/
+cd ${tmp_dir}/llvm-project/
 echo "Build the LLVM project"
 git checkout ${llvm_release}
 mkdir -p build
 cd build
 cmake -G "Ninja" -DLLVM_ENABLE_PROJECTS='clang;clang-tools-extra;libcxx;libcxxabi;libunwind;lld;lldb;compiler-rt;lld;polly;debuginfo-tests;openmp;parallel-libs' -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_CXX1Y=ON -DLLVM_ENABLE_EH=ON -DLLVM_ENABLE_RTTI=ON -DBUILD_SHARED_LIBS=ON -DLLVM_BUILD_EXAMPLES=Off -DLLVM_INCLUDE_EXAMPLES=Off -DLLVM_BUILD_TESTS=Off -DLLVM_INCLUDE_TESTS=Off -DPYTHON_EXECUTABLE=`which python3` ../llvm
```
vulder, line 38: If we reduce the list of build projects we can trim down the build time for users and reduce the possibility of compilation problems. I would assume, some of the listed project are/will never be used.
I'd remove: `lldb`, `polly`, `openmp`,`parallel-libs`

PS: `lld` is listed twice, once is enough :)

J03D03: True that. Thank you.

utils/install-llvm.sh
J03D03: Good point.

PR #155, opened 2020-06-21
Adds assert to trigger before analysis OOB access

vulder:


pdschubert:
Restructured the respective flow functions and added the suggested assert.

PR #170, opened 2020-07-23
Fix cmake calls in bootstrap.sh and install-llvm.sh

mailgerigk:
Seemed like a copy paste error in bootstrap.sh and install-llvm.sh was missing the "--" before the native options.


pdschubert:
Mhm, I just double checked on my machine and those commands run perfectly fine.
In the documentation on https://cmake.org/cmake/help/latest/command/build_command.html it states that
```
Sets the given <variable> to a command-line string of the form:

<cmake> --build . [--config <config>] [--target <target>...] [-- -i]
where <cmake> is the location of the cmake(1) command-line tool, and <config> and <target> are the values provided to the CONFIGURATION and TARGET options, if any. The trailing -- -i option is added for Makefile Generators if policy CMP0061 is not set to NEW.
```
What is the error message on your build system?

mailgerigk:
Running bootstrap.sh without the pull request results in 
```
Unknown argument -j8
Usage: cmake --build <dir> [options] [-- [native-options]]
Options:
  <dir>          = Project binary directory to be built.
  --target <tgt> = Build <tgt> instead of default targets.
                   May only be specified once.
  --config <cfg> = For multi-configuration tools, choose <cfg>.
  --clean-first  = Build target 'clean' first, then build.
                   (To clean only, use --target 'clean'.)
  --use-stderr   = Ignored.  Behavior is default in CMake >= 3.0.
  --             = Pass remaining options to the native tool.
```
coming from install-llvm.sh and
```
Unknown argument -j
Usage: cmake --build <dir> [options] [-- [native-options]]
Options:
  <dir>          = Project binary directory to be built.
  --target <tgt> = Build <tgt> instead of default targets.
                   May only be specified once.
  --config <cfg> = For multi-configuration tools, choose <cfg>.
  --clean-first  = Build target 'clean' first, then build.
                   (To clean only, use --target 'clean'.)
  --use-stderr   = Ignored.  Behavior is default in CMake >= 3.0.
  --             = Pass remaining options to the native tool.
```
for bootstrap.sh after fixing install-llvm.sh (note the missing number of threads in the 2nd error message because of the usage of num_cores instead of NUM_THREADS).

Tested on a fresh Ubuntu 18.04 only running
```
sudo apt update
sudo apt upgrade
sudo ./bootstrap.sh

cmake --version
>cmake version 3.10.2
```

MMory:
Hi @mailgerigk , sorry for leaving your PR open for that long. @pdschubert and myself just discussed the issue you're addressing with this PR.

As we are now using Ninja as the generator, there is no need for passing the -j flag at all as Ninja detects the number of jobs to be run in parallel on its own. Hence, if you alter your PR to remove the -j flag, we will merge it. We think that this would be the fairest solution to recognize and use your contribution.

PR #178, opened 2020-07-31
Implements EdgeFunctionSingletonFactory

vulder:
Some EdgeFunctions are allocated very often during analysis but only hold constant data, to reduce the memory consumption of these EdgeFunctions this commit introduces a SingletonFactory. With the SingletonFactory EdgeFunction implements can automatically reuse previously allocated EdgeFunctions that have the same internal data, saving large amounts of memory in cases where an EdgeFunction would be allocated many times.

### DIFF
--- a/include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
+++ b/include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
@@ -17,6 +17,7 @@
 #ifndef PHASAR_PHASARLLVM_IFDSIDE_EDGEFUNCTIONS_H_
 #define PHASAR_PHASARLLVM_IFDSIDE_EDGEFUNCTIONS_H_
 
+#include "llvm/Config/llvm-config.h"
 #include "llvm/Support/Compiler.h"
 
 #include <atomic>
@@ -585,8 +586,10 @@ private:
         : CleanerThread(&EdgeFactoryStorageCleaner::runCleaner, this,
                         std::reference_wrapper<EFStorageData>(
                             EdgeFunctionSingletonFactory::getCacheData())) {
+#ifdef LLVM_ON_UNIX
       pthread_setname_np(CleanerThread.native_handle(),
                          "EdgeFactoryStorageCleanerThread");
+#endif
     }
     EdgeFactoryStorageCleaner(const EdgeFactoryStorageCleaner &) = delete;
     EdgeFactoryStorageCleaner &
--- a/include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
+++ b/include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
@@ -17,13 +17,20 @@
 #ifndef PHASAR_PHASARLLVM_IFDSIDE_EDGEFUNCTIONS_H_
 #define PHASAR_PHASARLLVM_IFDSIDE_EDGEFUNCTIONS_H_
 
+#include "llvm/Support/Compiler.h"
+
+#include <atomic>
 #include <iosfwd>
 #include <iostream>
+#include <map>
 #include <memory>
+#include <mutex>
 #include <ostream>
 #include <set>
 #include <sstream>
 #include <string>
+#include <thread>
+#include <utility>
 
 namespace psr {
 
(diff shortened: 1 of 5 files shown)


include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
```diff
@@ -472,6 +482,149 @@ template <typename AnalysisDomainTy> class EdgeFunctions {
...
+// EdgeFunction that have constant data only need to be allocated once and can
+// be turned into a singleton, which save large amounts of memory. The
+// automatic singleton creation can be enbaled for a EdgeFunction class by
+// inheriting from EdgeFunctionSingletonFactory and only allocating
+// EdgeFunction throught the provided createEdgeFunction method.
+//
+// Old unused EdgeFunction that were deallocated need to/should be clean from
+// the factory by either calling cleanExpiredEdgeFunctions or initializing the
+// automatic cleaner thread with initEdgeFunctionCleaner.
+template <typename EdgeFunctionType, typename CtorArgT>
```
MMory, line 497: @pdschubert we should discuss template parameter naming to achieve consistency within phasar.

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
```diff
@@ -472,6 +482,149 @@ template <typename AnalysisDomainTy> class EdgeFunctions {
...
+  EdgeFunctionSingletonFactory &
+  operator=(EdgeFunctionSingletonFactory &&) noexcept = default;
+
+  virtual ~EdgeFunctionSingletonFactory() {
+    std::lock_guard<std::mutex> DataLock(getCacheData().DataMutex);
+  }
+
+  // Creates a new EdgeFunction of type EdgeFunctionType, reusing the previous
+  // allocation if an EdgeFunction with the same values was already created.
+  static inline std::shared_ptr<EdgeFunctionType>
```
MMory, line 515: shared_ptr again?

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
```diff
@@ -472,6 +482,149 @@ template <typename AnalysisDomainTy> class EdgeFunctions {
...
+    }
+  }
+
+  class EdgeFactoryStorageCleaner {
+  public:
+    EdgeFactoryStorageCleaner()
+        : CleanerThread(&EdgeFactoryStorageCleaner::runCleaner, this,
+                        std::reference_wrapper<EFStorageData>(
+                            EdgeFunctionSingletonFactory::getCacheData())) {
+      pthread_setname_np(CleanerThread.native_handle(),
```
MMory, line 588: the windows counterpart could be https://stackoverflow.com/a/23899379
honestly, that windows code looks pretty ugly.

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
vulder: As long as we have no other memory management in phasar we need to stick to the shared_ptr approach. If we get the memory manager working, the factory can be updated to work with the mem manager.

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
vulder: An easy alternative for windows could also be to just remove the thread naming. Makes it a bit less convenient for windows users but saves us some trouble.

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
pdschubert: EdgeFunctionType is fine here. We consistently use that type name for the flow/edge function memory manager, too.

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/EdgeFunctions.h
pdschubert: According to the pthread_setname_np man page it states
"""
The pthread_setname_np() function can be used to  set  a
       unique  name  for  a  thread,  which can be useful for debugging multi‐
       threaded applications.  The thread name  is  a  meaningful  C  language
       string, whose length is restricted to 16 characters, including the ter‐
       minating null byte ('\0').
"""
But anyways, I think we could get rid of thread naming.

vulder: You want to fully remove it? It's quite handy for unix envs.

vulder: I added ifdefs so only unix systems use thread naming

PR #180, opened 2020-08-12
fix search for boost when using phasar as library

MMory:
On ubuntu 20 using phasar as a library was broken:

CMake Error at CMakeLists.txt:12 (add_executable):
 Target "myphasartool" links to target "Boost::graph" but the target was not
 found. Perhaps a find_package() call is missing for an IMPORTED target, or
 an ALIAS target is missing?
…

this PR should fix this by triggering searches for the dependent boost components through phasar_*-config.cmake.

@vulder Please check whether this breaks the in-tree build in vara. I was unable to build vara with the current state of the phasar development branch.

### DIFF
--- a/lib/Config/CMakeLists.txt
+++ b/lib/Config/CMakeLists.txt
@@ -16,8 +16,11 @@ else()
   )
 endif()
 
+find_package(Boost COMPONENTS filesystem REQUIRED)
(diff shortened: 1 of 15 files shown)


vulder:
works for me

PR #182, opened 2020-08-19
Add ability to get a set of all functions in the callgraph.

ddiepo-pjr:
An existing function, getDependencyOrderedFunctions(), does have the
ability to get all of the functions from the callgraph. However, this
does extra work to walk the graph such that the functions are returned
in a particular order.  This new function returns the functions in a set
so that we can quickly identify if a given function from the input is
within the graph.  A flat set was chosen for efficiency of space and
speed of lookups.


pdschubert (approved):
Thanks!

PR #185, opened 2020-08-26
add gerneal cmake --build option (over 3.0 version)

dbwodlf3:
cmake --build . -j${num_cores}
==>
cmake --build . -- -j${num_cores}

cmake --build -j option is available over 3.12

but, -- -j option is available over 3.0

i think it is better.

https://stackoverflow.com/questions/36633074/set-the-number-of-threads-in-a-cmake-build
https://cmake.org/cmake/help/v3.0/manual/cmake.1.html


dbwodlf3:
(now, 2020.08.26) Ubuntu 18.04 cmake version(apt-get install cmake) is 3.10.2.

dbwodlf3:
same pr

PR #186, opened 2020-08-28
Makes memory improvements in FlowEdgeFunctionCache

vulder:
Improves memory consumption by compressing keys of maps under a huge load and switching certain maps to a new EquivalenceClassMap, which only keeps one representative value for each equivalence class of keys.
Furthermore, unnecessary count checks were removed by direct look up of the keys, which saves one map traversal for every look up.


pdschubert (approved):
Looks alright, builds fine, and all unit tests pass. Thanks!

PR #187, opened 2020-08-29
Using correct bash variable for threads

Machiry:
Fixing the variable name that contains the number of threads.


pdschubert:
Thanks!

PR #196, opened 2020-09-15
Small speed improvements for BitVector and Table

vulder:
Add branch likleyhood information because the condition will only be
true once, in the case where we exit the function.


pdschubert (approved):
Looks good. Thanks!

PR #205, opened 2020-10-13
Fix Assertion Violation

Divyank004:
Added the necessary checks to prevent the assertion check fail.


pdschubert (approved):
Looks alright. Thanks Divyank.

PR #207, opened 2020-10-17
Update bootstrap script (set ENV variables & shorten arch installation)

J03D03:
# Environment variables
Add a bash script to the `utils` folder which sets the environment variables after the PhASAR installation. It appends the export commands to the `~/.$(basename $SHELL)rc` file.

# Arch update
Install (`--needed`) packages if they are not already installed.


pdschubert (approved):
Looks good. Thanks.

Issue #206, opened 2020-10-16
Missing callgraph functions with FFTW library


btruong-pjr:
- [X] I have searched open and closed issues for duplicates
  - [X] I made sure that I am not using an old project version (DO: pull Phasar, update git submodules, rebuild the project and check if the bug is still there)

----------------------------------------

### Bug description
There are missing functions from the callgraph when using libfftw.

### Steps to reproduce
  - Compile libfftw-3.3.7
  - Compile sample fftw program
  - Generate callgraph
  - Verify callgraph does not contain missing functions
**Actual result:** I ran a dynamic analysis to generate a callgraph of what functions were actually being called. The results are in 
[dynamic_functions.txt](https://github.com/secure-software-engineering/phasar/files/5394685/dynamic_functions.txt)

**Expected result:** Attached is the callgraph that shows these functions as missing:
[callgraph.log](https://github.com/secure-software-engineering/phasar/files/5394687/callgraph.log)
apply_cpy2dco
apply_ip_sq
apply_ip_sq_tiled
apply_ip_sq_tiledbuf
apply_iter
apply_tiled
apply_tiledbuf
compare_by_istride
copy
dotile
dotile
dotile_buf
dotile_buf
fftwf_cpy1d
fftwf_cpy2d_tiled
fftwf_cpy2d_tiledbuf
fftwf_dft_solve
fftwf_null_awake
fftwf_rdft_solve
fftwf_tile2d
fftwf_transpose
fftwf_transpose_tiled
fftwf_transpose_tiledbuf
okp
okp
transpose
transpose_rec

### Context (Environment)

  - **phasar:** \[cd94f1a48d05161c44e166074e94be90f95d2c6c]
  - **googletest:** \[042e412c36880afc882bebfe3f79024bf7aa24b7]
  - **json:** \[042e412c36880afc882bebfe3f79024bf7aa24b7]
  - **WALi-OpenNWA:** \[73f9d7a03367b2c485cec35c4bea0d02e7e87742]

**Operating System:**
  - [X] Linux
  - [ ] Windows
  - [ ] macOS

**Build Type:** 
  - [X] cmake
  - [ ] custom build

### Possible solution

We are happy to discuss possible solutions to this problem, especially if it origniates from a design flaw.

### Example files

**Files:**
[fftw_sample.tar.gz](https://github.com/secure-software-engineering/phasar/files/5411932/fftw_sample.tar.gz)
[...]

PR #212, opened 2020-10-28
Fix C-style polymorphism

MMory:
This fixes several bugs that lead to potentially incomplete callgraphs, see https://github.com/secure-software-engineering/phasar/issues/206

Please review and merge https://github.com/secure-software-engineering/phasar/pull/211 first as callgraph dot export benefits greatly from that as well.

### DIFF
--- a/include/phasar/PhasarLLVM/ControlFlow/LLVMBasedICFG.h
+++ b/include/phasar/PhasarLLVM/ControlFlow/LLVMBasedICFG.h
@@ -25,6 +25,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
+#include <stack>
 
 #include "boost/container/flat_set.hpp"
 #include "boost/graph/adjacency_list.hpp"
@@ -72,6 +73,12 @@ private:
   // Any types that could be initialized outside of the module
   // std::set<const llvm::StructType*> unsound_types;
 
+  // The worklist for direct callee resolution.
+  std::stack<const llvm::Function *> FunctionWL;
+
+  // Map indirect calls to the number of possible targets found for it. Fixpoint
+  // is not reached when more targets are found.
+  std::unordered_map<const llvm::Instruction *, unsigned> IndirectCalls;
   // The VertexProperties for our call-graph.
   struct VertexProperties {
     const llvm::Function *F = nullptr;
@@ -107,7 +114,10 @@ private:
   /// Maps functions to the corresponding vertex id.
   std::unordered_map<const llvm::Function *, vertex_t> FunctionVertexMap;
 
-  void constructionWalker(const llvm::Function *F, Resolver &Resolver);
+  void processFunction(const llvm::Function *F, Resolver &Resolver,
+                           bool &FixpointReached);
+
+  bool constructDynamicCall(const llvm::Instruction *I, Resolver &Resolver);
 
   std::unique_ptr<Resolver> makeResolver(ProjectIRDB &IRDB,
                                          CallGraphAnalysisType CGT,
--- a/include/phasar/PhasarLLVM/ControlFlow/LLVMBasedICFG.h
+++ b/include/phasar/PhasarLLVM/ControlFlow/LLVMBasedICFG.h
@@ -21,11 +21,11 @@
 #include <iostream>
 #include <memory>
 #include <set>
+#include <stack>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
-#include <stack>
 
 #include "boost/container/flat_set.hpp"
 #include "boost/graph/adjacency_list.hpp"
@@ -115,7 +115,7 @@ private:
   std::unordered_map<const llvm::Function *, vertex_t> FunctionVertexMap;
 
   void processFunction(const llvm::Function *F, Resolver &Resolver,
-                           bool &FixpointReached);
+                       bool &FixpointReached);
 
   bool constructDynamicCall(const llvm::Instruction *I, Resolver &Resolver);
 
--- a/lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
+++ b/lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
@@ -256,14 +256,14 @@ bool LLVMBasedICFG::constructDynamicCall(const llvm::Instruction *I,
   bool NewTargetsFound = false;
   // Find vertex of calling function.
   vertex_t ThisFunctionVertexDescriptor;
-  auto FvmItr = FunctionVertexMap.find(I->getParent()->getParent());
+  auto FvmItr = FunctionVertexMap.find(I->getFunction());
   if (FvmItr != FunctionVertexMap.end()) {
     ThisFunctionVertexDescriptor = FvmItr->second;
   } else {
     LOG_IF_ENABLE(
         BOOST_LOG_SEV(lg::get(), ERROR)
         << "constructDynamicCall: Did not find vertex of calling function "
-        << I->getParent()->getParent()->getName().str() << " at callsite "
+        << I->getFunction()->getName().str() << " at callsite "
         << llvmIRToString(I));
     std::terminate();
   }
--- a/lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
+++ b/lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
@@ -118,7 +118,26 @@ LLVMBasedICFG::LLVMBasedICFG(ProjectIRDB &IRDB, CallGraphAnalysisType CGType,
     if (F == nullptr) {
       llvm::report_fatal_error("Could not retrieve function for entry point");
     }
-    constructionWalker(F, *Res);
+    FunctionWL.push(F);
+  }
+  bool FixpointReached;
+  do {
+    FixpointReached = true;
+    while (!FunctionWL.empty()) {
+      const llvm::Function *F = FunctionWL.top();
+      FunctionWL.pop();
+      processFunction(F, *Res, FixpointReached);
+    }
+    for (const auto &[Callsite, _] : IndirectCalls) {
+      FixpointReached &= !constructDynamicCall(Callsite, *Res);
+    }
+  } while (!FixpointReached);
+  for (const auto &[IndirectCall, Targets] : IndirectCalls) {
+    if (Targets == 0) {
+      LOG_IF_ENABLE(BOOST_LOG_SEV(lg::get(), WARNING)
+                    << "No callees found for callsite "
+                    << llvmIRToString(IndirectCall));
+    }
   }
   REG_COUNTER("CG Vertices", getNumOfVertices(), PAMM_SEVERITY_LEVEL::Full);
   REG_COUNTER("CG Edges", getNumOfEdges(), PAMM_SEVERITY_LEVEL::Full);
@@ -137,8 +156,8 @@ LLVMBasedICFG::~LLVMBasedICFG() {
   }
 }
 
-void LLVMBasedICFG::constructionWalker(const llvm::Function *F,
-                                       Resolver &Resolver) {
+void LLVMBasedICFG::processFunction(const llvm::Function *F, Resolver &Resolver,
+                                    bool &FixpointReached) {
   PAMM_GET_INSTANCE;
   LOG_IF_ENABLE(BOOST_LOG_SEV(lg::get(), DEBUG)
                 << "Walking in function: " << F->getName().str());
@@ -191,12 +210,9 @@ void LLVMBasedICFG::constructionWalker(const llvm::Function *F,
                           << "Found dynamic call-site: ");
             LOG_IF_ENABLE(BOOST_LOG_SEV(lg::get(), DEBUG)
                           << "  " << llvmIRToString(CS.getInstruction()));
-            // call the resolve routine
-            if (LLVMBasedICFG::isVirtualFunctionCall(CS.getInstruction())) {
-              PossibleTargets = Resolver.resolveVirtualCall(CS);
-            } else {
-              PossibleTargets = Resolver.resolveFunctionPointer(CS);
-            }
+            IndirectCalls[&I] = 0;
+            FixpointReached = false;
+            continue;
           }
         }
 
@@ -223,7 +239,7 @@ void LLVMBasedICFG::constructionWalker(const llvm::Function *F,
 
         // continue resolving
         for (const auto *PossibleTarget : PossibleTargets) {
-          constructionWalker(PossibleTarget, Resolver);
+          FunctionWL.push(PossibleTarget);
         }
 
         Resolver.postCall(&I);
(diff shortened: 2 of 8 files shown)


MMory:
@pdschubert the fix breaks a test in the constness analysis, let's discuss that. I am not sure whether the test was correct in the first place. Should be caused by the changed resolution of pts of globals.

include/phasar/PhasarLLVM/ControlFlow/LLVMBasedICFG.h
```diff
@@ -107,7 +114,10 @@ class LLVMBasedICFG
   /// Maps functions to the corresponding vertex id.
   std::unordered_map<const llvm::Function *, vertex_t> FunctionVertexMap;
 
-  void constructionWalker(const llvm::Function *F, Resolver &Resolver);
+  void processFunction(const llvm::Function *F, Resolver &Resolver,
+                           bool &FixpointReached);
```
vulder, line 118: Formatting ?

lib/Utils/LLVMShorthands.cpp
```diff
@@ -66,20 +66,60 @@ bool isAllocaInstOrHeapAllocaFunction(const llvm::Value *V) noexcept {
...
+        }
+        ++Idx;
+      }
+    } else {
+      for (const auto &Arg : F->args()) {
+        if (!isTypeMatchForFunctionArgument(FType->getParamType(Idx),
+                                            Arg.getType())) {
+          return false;
+        }
+        ++Idx;
```
vulder, line 121:
```suggestion
    for (const auto &Arg : F->args()) {
      bool TypeMissMatch = ExactMatch ? Arg.getType() != FType->getParamType(Idx) : 
                                                                  !isTypeMatchForFunctionArgument(FType->getParamType(Idx), Arg.getType());
      if (TypeMissMatch) {
        return false;
      }
      ++Idx;
    }
```
I think this is a bit easier to understand and does not duplicate the loop code.

PS: code needs formatting, the gh platform is not the best for editing code :wink:

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
```diff
@@ -144,18 +140,73 @@ OTFResolver::resolveFunctionPointer(llvm::ImmutableCallSite CS) {
...
+            if (auto *CA = llvm::dyn_cast<llvm::ConstantAggregate>(Op)) {
+              ConstantAggregateWL.push(CA);
+            }
+            if (auto *GV = llvm::dyn_cast<llvm::GlobalVariable>(Op)) {
+              if (GV->hasInitializer()) {
+                if (auto *GVCA = llvm::dyn_cast<llvm::ConstantAggregate>(
+                        GV->getInitializer())) {
+                  ConstantAggregateWL.push(GVCA);
+                }
+              }
```
vulder, line 203:
```suggestion
              if (!GV->hasInitializer()) {
                continue;
              }
              if (auto *GVCA = llvm::dyn_cast<llvm::ConstantAggregate>(
                      GV->getInitializer())) {
                ConstantAggregateWL.push(GVCA);
              }
```
If I understand the code correctly, we could use a `continue` to reduces the nested if structure a bit.

vulder (commented):
I made a quick pass over the code :) and added two suggestions.

include/phasar/PhasarLLVM/ControlFlow/LLVMBasedICFG.h
MMory: @pdschubert will do this at the time of merge. There is other unrelated stuff where formatting needs to be done.

lib/Utils/LLVMShorthands.cpp
MMory: Agreed, the compiler should be smart enough to find that ExactMatch is loop invariant and do the duplication I did in source code.

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
MMory: Agreed.

lib/Utils/LLVMShorthands.cpp
```diff
@@ -76,10 +106,14 @@ bool matchesSignature(const llvm::Function *F,
...
     for (const auto &Arg : F->args()) {
-      if (Arg.getType() != FType->getParamType(Idx)) {
+      bool TypeMissMatch = ExactMatch ? Arg.getType() != FType->getParamType(Idx) : 
+                                                                  !isTypeMatchForFunctionArgument(FType->getParamType(Idx), Arg.getType());
+      if (TypeMissMatch) {
         return false;
       }
       ++Idx;
+    }
+      }
```
ddiepo, line 116: These two closing braces are extraneous and cause compilation errors.

I'm trying the github suggestion feature, but I'm not sure how that will work in this case:
```suggestion
```

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
```diff
@@ -144,18 +140,74 @@ OTFResolver::resolveFunctionPointer(llvm::ImmutableCallSite CS) {
...
+            if (auto *CA = llvm::dyn_cast<llvm::ConstantAggregate>(Op)) {
+              ConstantAggregateWL.push(CA);
+            }
+            if (auto *GV = llvm::dyn_cast<llvm::GlobalVariable>(Op)) {
+              if (!GV->hasInitializer()) {
+                continue;
+              }
+              if (auto *GVCA = llvm::dyn_cast<llvm::ConstantAggregate>(
+                      GV->getInitializer())) {
+                ConstantAggregateWL.push(GVCA);
```
ddiepo, line 203: I believe there's a possibility for an infinite loop here with the pushing things onto the ConstantAggregateWL.

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
```diff
@@ -144,18 +140,74 @@ OTFResolver::resolveFunctionPointer(llvm::ImmutableCallSite CS) {
...
+          }
+          auto InitConst = GV->getInitializer();
+          if (auto *InitConstAggregate =
+                  llvm::dyn_cast<llvm::ConstantAggregate>(InitConst)) {
+            ConstantAggregateWL.push(InitConstAggregate);
+          }
+        }
+        while (!ConstantAggregateWL.empty()) {
+          auto ConstAggregateItem = ConstantAggregateWL.top();
+          ConstantAggregateWL.pop();
```
ddiepo, line 175: See next comment about possible infinite loop:  You can avoid it by keeping track of the worklist items and avoid re-processing.  I admit, I don't fully understand what's going on here, so please review closely!  I can say that without this change, I have a sample program which appear to fall into an infinite loop in this section of code.
```suggestion
        std::set<const llvm::ConstantAggregate *> VisitedConstantAggregates;
        while (!ConstantAggregateWL.empty()) {
          auto ConstAggregateItem = ConstantAggregateWL.top();
          ConstantAggregateWL.pop();
          // We may have already processed the item, avoid an infinite loop:
          if (!VisitedConstantAggregates.insert(ConstAggregateItem).second)
            continue;
```

lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
```diff
@@ -234,6 +250,85 @@ void LLVMBasedICFG::constructionWalker(const llvm::Function *F,
...
+        FunctionVertexMap[PossibleTarget] = TargetVertex;
+      }
+      boost::add_edge(ThisFunctionVertexDescriptor, TargetVertex,
+                      EdgeProperties(I), CallGraph);
+    }
+
+    // continue resolving
+    for (const auto *PossibleTarget : PossibleTargets) {
+      FunctionWL.push(PossibleTarget);
+    }
```
ddiepo, line 323: Can you skip adding the function to the worklist if it has already been processed?
```suggestion
    // continue resolving
    for (const auto *PossibleTarget : PossibleTargets) {
      if (VisitedFunctions.count(PossibleTarget) == 0) {
          FunctionWL.push(PossibleTarget);
      }
    }
```

ddiepo (commented):
I've been running this branch against a sample program and it was getting hung up without the suggested change in OTFResolver.  Now it runs to the end, but there appears to be some final cleanup errors, as I'm seeing a "stack smashing detected" core dump upon completion now.  I haven't yet researched what's going on there -- it could be in my wrapper.

lib/Utils/LLVMShorthands.cpp
vulder: Oh, that was probably my bad and got tangled up with my suggestions. Getting the correct nesting depth and indentation of brackets right with github review suggestions is sometimes hard :laughing:

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
MMory: Could you provide your sample program please? This part here is supposed to recurse into nested global objects, which should be finite.

lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
MMory: Nope. The points-to information could have been updated, leading to new call edges which were not seen before.

ddiepo-pjr: That makes sense.  This turned out to not be a problem for the program I was analyzing.

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
ddiepo-pjr: Oops, I used the wrong github account on my initial review!
I'll have to create sample, the program I was analyzing is not something I have the ability to share.  But I can say what I saw from GDB was the variable "GVCA" at line 203 was the same as "ConstAggregateItem" from line 174, thus the logic would not exit the while loop since ConstantAggregateWL never becomes empty.  I can look into this more in a few days and give you more specifics and/or some sample code.

lib/Utils/LLVMShorthands.cpp
ddiepo-pjr: Yes!  The suggestions are a bit tricky since it does not show the surrounding code.

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
MMory: It would suffice to isolate the data structure where this occurs. That would show how a ConstantAggregate can have itself as an initializer of its own operand, which seems very weird to me.

ddiepo-pjr: Good news, I was able to track down where this was happening in the application this morning and it is in an open source library.

Super short version -- I believe this is the struct in question:
```
struct metal_list {
	struct metal_list *next, *prev;
};
```

In https://github.com/OpenAMP/libmetal/blob/master/lib/irq.c  the variable irq_cntrs which is declared as part of the macro on line 14.  That macro is in https://github.com/OpenAMP/libmetal/blob/master/lib/list.h line 38/39, with the related code from lines 25-33.

Here's some gdb output that may be helpful:

```
(gdb) bt
#0  psr::OTFResolver::resolveFunctionPointer (this=0x5555565289c0, CS=...) at ../lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp:206
#1  0x00007ffff7e59e2a in psr::LLVMBasedICFG::constructDynamicCall (this=0x7fffffffdf28, I=0x555556554cf0, Resolver=...) at ../lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp:285
#2  0x00007ffff7e5b093 in psr::LLVMBasedICFG::LLVMBasedICFG (this=0x7fffffffdf28, IRDB=..., CGType=psr::CallGraphAnalysisType::OTF, EntryPoints=std::set with 1 element = {...}, TH=0x7fffffffdd90, 
    PT=0x55555655b920, SF=psr::SoundnessFlag::SOUNDY) at ../lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp:134
#3  0x000055555558ac08 in ArtusPhasar::CallgraphContainer::CallgraphContainer(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, psr::IRDBOptions, psr::CallGraphAnalysisType, std::set<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) ()
#4  0x000055555556c5d8 in main ()
Breakpoint 1, psr::OTFResolver::resolveFunctionPointer (this=0x5555565289c0, CS=...) at ../lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp:206
206                     ConstantAggregateWL.push(GVCA);
(gdb) print GVCA
$1 = (llvm::ConstantAggregate *) 0x5555556ad500
(gdb) print ConstAggregateItem
$2 = (const llvm::ConstantAggregate *) 0x5555556ad500
(gdb) print *GV
$5 = {<llvm::GlobalObject> = {<llvm::GlobalValue> = {<llvm::Constant> = {<llvm::User> = {<llvm::Value> = {VTy = 0x5555555ff940, UseList = 0x55555657d870, SubclassID = 3 '\003', HasValueHandle = 0 '\000', 
            SubclassOptionalData = 0 '\000', SubclassData = 0, NumUserOperands = 1, IsUsedByMD = 0, HasName = 1, HasHungOffUses = 0, HasDescriptor = 0, static MaxAlignmentExponent = 29, 
            static MaximumAlignment = 536870912}, <No data fields>}, <No data fields>}, ValueType = 0x5555555fd190, static GlobalValueSubClassDataBits = 16, Linkage = 7, Visibility = 0, UnnamedAddrVal = 0, 
      DllStorageClass = 0, ThreadLocal = 0, HasLLVMReservedName = 0, IsDSOLocal = 1, HasPartition = 0, SubClassData = 36, IntID = 0, Parent = 0x5555555fa180}, ObjComdat = 0x0, 
    static GlobalObjectSubClassDataBits = 9, static AlignmentBits = 5, static AlignmentMask = 31, 
    static GlobalObjectMask = 127}, <llvm::ilist_node<llvm::GlobalVariable>> = {<llvm::ilist_node_impl<llvm::ilist_detail::node_options<llvm::GlobalVariable, false, false, void> >> = {<llvm::ilist_node_base<false>> = {Prev = 0x5555556ad3a0, Next = 0x5555555fa188}, <No data fields>}, <No data fields>}, Attrs = {SetNode = 0x0}, isConstantGlobal = false, isExternallyInitializedConstant = false}
(gdb) print GV->getName()
$6 = {static npos = 18446744073709551615, Data = 0x5555556ad490 "irq_cntrs", Length = 9}
(gdb) c
Continuing.

Breakpoint 1, psr::OTFResolver::resolveFunctionPointer (this=0x5555565289c0, CS=...) at ../lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp:206
206                     ConstantAggregateWL.push(GVCA);
(gdb) print GVCA
$7 = (llvm::ConstantAggregate *) 0x5555556ad500
(gdb) print ConstantAggregateWL
$9 = std::stack wrapping: std::deque with 2593485 elements = {0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 
  0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500, 0x5555556ad500...}
```

ddiepo-pjr: Also, I found that the test app that comes with libmetal has the same issue as above, and can be used to reproduce the problem with irq_cntrs.  libmetal can be built with wllvm, and the extract the .ll file from test-metal-static for use.

MMory: Thank you, David, I overlooked this case. I'll look into this again.

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
MMory: I just pushed a fix that is almost identical to your suggestion. Thanks again!

ddiepo-pjr (commented):
Good improvements from my suggestion.  Thanks!

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
ddiepo-pjr: I verified the application analysis does not get stuck in an infinite loop now with the latest version.  Thanks!

pdschubert (approved):
Looks good. Thanks, all of you.

Issue #194, opened 2020-09-14
 Assertion `!isLiteral() && "Literal structs never have names"' failed.


legendtang:
- [x] I have searched open and closed issues for duplicates
  - [x] I made sure that I am not using an old project version (DO: pull Phasar, update git submodules, rebuild the project and check if the bug is still there)

----------------------------------------

### Bug description
When I try to analyze a bitcode built from SPEC 2017,
phasar throws an assertion failure

`phasar-llvm: /home/username/llvm10/llvm/lib/IR/Type.cpp:505: llvm::StringRef llvm::StructType::getName() const: Assertion `!isLiteral() && "Literal structs never have names"' failed.`

### Steps to reproduce
run the below bitcode with any modes.

### Context (Environment)

Phasar commit ID: 82733ae
OS: CentOS 7
LLLVM version: 10.0.1

**Build Type:** 
  - [x] cmake
  - [ ] custom build

### Possible solution

Add some checks to it.

### Example files

**Files:**

https://mega.nz/file/vZ4FlAYK#z_RUrsUz_3Ljfll67VGG-IcxiDSFl-t1O3e0uUWtUao

pdschubert:
Thanks for your report.

I tried to reproduce the issue using the following commands:
```
$ ulimit -s 16777216
$ tools/phasar-llvm/phasar-llvm -m cpugcc_r_base.mytest-m64.bc -C CHA -D ifds-solvertest
```
using a debug compile of Phasar based on the latest commit on the `development` branch.

However, I've not been able to reproduce the issue and the analysis terminated normally.
Could you check if the latest state on `development` also fixes the bug for you?

legendtang:
> Thanks for your report.
> 
> I tried to reproduce the issue using the following commands:
> 
> ```
> $ ulimit -s 16777216
> $ tools/phasar-llvm/phasar-llvm -m cpugcc_r_base.mytest-m64.bc -C CHA -D ifds-solvertest
> ```
> 
> using a debug compile of Phasar based on the latest commit on the `development` branch.
> 
> However, I've not been able to reproduce the issue and the analysis terminated normally.
> Could you check if the latest state on `development` also fixes the bug for you?

Sorry for the misinformation I provided but the bug still exists with DTA even in the latest dev branch. We can use a smaller bc to test it out:

`phasar-llvm -C DTA -m leela_r_base.mytest-m64.bc`

[...]

PR #213, opened 2020-10-29
Fix https://github.com/secure-software-engineering/phasar/issues/194

MMory:
Issue can be reproduced with debug build of LLVM. This fixes the issue by handling literal structs properly.


pdschubert (approved):
Thanks!

PR #215, opened 2020-10-30
Add assert to guarantee that D1FactId is not a nullptr

vulder:
Clang static analyze finds the issue that D1FactId could be nullptr,
however, the code looks fine. Hence, we place an assert here to make
sure this case never actually happens during an analysis run.


pdschubert (approved):
Thanks.

PR #217, opened 2020-11-02
Add assert to guarantee that D1_FSG is not a nullptr

vulder:
Clang static analyze finds the issue that D1_FSG could be nullptr,
however, the code looks fine. Hence, we place an assert here to make
sure this case never actually happens during an analysis run.


MMory (approved):
In the current code the assert should never be hit. I agree to insert the assert to silence the nullness finding.

pdschubert (approved):
Okay, that should work now.

PR #218, opened 2020-11-05
Fixed boundary-value-bug of erase() in BitVectorSet.h

drop-stones:
I found a boundary-value-bug at l.261 of erase() in BitVectorSet.h.
In this code, we cannot erase the first element because of underflow.
So change the condition check to the same as that at l.278 of count() in BitVectorSet.h.

Please review this change.


pdschubert (approved):
Thanks.

PR #222, opened 2020-12-08
f-AdvancedIIATesting

pdschubert:


include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
```diff
@@ -670,9 +734,56 @@ class IDEInstInteractionAnalysisT
...
+    l_t UserEdgeFacts = BitVectorSet<e_t>();
+    std::set<e_t> EdgeFacts;
+    if (edgeFactGen) {
+      EdgeFacts = edgeFactGen(callSite);
+      // fill BitVectorSet
+      UserEdgeFacts = BitVectorSet<e_t>(EdgeFacts.begin(), EdgeFacts.end());
+    }
+    // Model call to heap allocating functions (new, new[], malloc, etc.) --
+    // only model direct calls, though.
+    if (callees.size() == 1) {
```
vulder, line 747: What happens in the case we have non direct calls, maybe through an allocation? Not sure if this is correctly handled by the other code.

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
```diff
@@ -253,26 +261,28 @@ class IDEInstInteractionAnalysisT
...
-              Facts.erase(src);
-            }
+            // If the value to be stored does not hold we must at least add the
+            // store instruction and the points-to set as the instruction still
+            // interacts with the memory locations pointed to be PTS.
+            // if (Store->getPointerOperand() == src ||
+            //        PointerPTS->count(src)) {
+            //   Facts.insert(Store);
+            //  Facts.erase(src);
+            // }
```
vulder, line 275: Why is this commented? Not needed or did you miss replacing it?

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
```diff
@@ -147,30 +150,32 @@ class IDEInstInteractionAnalysisT
...
+    //           // this->indirecrTaints[succ].insert(src);
+    //           for (auto &inst : succ->instructionsWithoutDebug()) {
+    //             ret.insert(&inst);
+    //           }
+    //         }
+    //       }
+    //       return ret;
+    //     });
+    //   }
+    // }
```
vulder, line 171: Hmm, I think this should be readded or am I wrong? Or do we handle the propagation of indirect taints somewhere else?

vulder (changes requested):
Found just a few things we should clean up before merging.
:+1: for the comments, makes it much easier to follow what the analysis should be doing.

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
```diff
@@ -936,28 +1047,38 @@ class IDEInstInteractionAnalysisT
...
-    if (Lhs == BottomElement || Rhs == BottomElement) {
-      return BottomElement;
-    }
     if (Lhs == TopElement) {
       return Rhs;
     }
     if (Rhs == TopElement) {
       return Lhs;
     }
+    if (Lhs == BottomElement) {
```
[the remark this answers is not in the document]
pdschubert: I don't have hard feelings about that. Will resort to the `or` variant.

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
pdschubert: Let's discuss that in our next meeting.

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
pdschubert: Added it again. There is also a bug in the code which I fixed on the fly.

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
pdschubert: I need to double check on that one. There was some issue here ...

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
```diff
@@ -146,25 +146,48 @@ class IDEInstInteractionAnalysisT
...
-    //           }
-    //         }
-    //       }
-    //       return ret;
-    //     });
-    //   }
-    // }
+    // Handle indirect taints, i. e., propagate values that depend on branch
+    // conditions whose operands are tainted.
+    if (EnableIndirectTaints) {
```
vulder, line 151:
```suggestion
    if constexpr (EnableIndirectTaints) {
```
Should not matter much but putting constexpr here makes it clear that EnableIndirectTaints is a compile time constant and the block will not be compiled in when it is `false`

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
```diff
@@ -146,25 +146,48 @@ class IDEInstInteractionAnalysisT
...
+      if (auto Br = llvm::dyn_cast<llvm::BranchInst>(curr);
```
vulder, line 152:
```suggestion
     if (const auto *Br = llvm::dyn_cast<llvm::BranchInst>(curr);
```

vulder (changes requested):

vulder:
@pdschubert 
One thing I missed, I think we should extend the tests for the analysis to capture the problems we found so we can prevent them from being reintroduced.

vulder (approved):
Fixes work for my example codes :+1:

PR #224, opened 2020-12-11
Fix a few minor issues

benquike:
* An error in ExampleModulePass (see code change)
* Fix a few memory leaks in unittests (some objects are not deleted in teardown routines)

### DIFF
--- a/unittests/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IFDSUninitializedVariablesTest.cpp
+++ b/unittests/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IFDSUninitializedVariablesTest.cpp
@@ -1,3 +1,5 @@
+#include <memory>
+
 #include "phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IFDSUninitializedVariables.h"
 #include "phasar/DB/ProjectIRDB.h"
 #include "phasar/PhasarLLVM/ControlFlow/LLVMBasedICFG.h"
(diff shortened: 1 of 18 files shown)


unittests/TestUtils/TestConfig.h
```diff
@@ -2,10 +2,11 @@
 #define UNITTEST_TESTUTILS_TESTCONFIG_H_
 
 #include <string>
+#include "config.h"
```
vulder, line 5:
```suggestion
#include "config.h"

#include <string>
```

unittests/PhasarLLVM/DataFlowSolver/Mono/InterMonoTaintAnalysisTest.cpp
```diff
@@ -77,6 +78,8 @@ class InterMonoTaintAnalysisTest : public ::testing::Test {
       FoundResults.insert(getMetaDataID(Result));
     }
     EXPECT_EQ(FoundResults, GroundTruth);
+
+    delete PT;
```
vulder, line 82: Fixing the leak here is a good idea :smile: , however, I think we should just replace the call to `new` with a `make_unique` and use `unique_ptr` instead.

unittests/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IFDSUninitializedVariablesTest.cpp
```diff
@@ -49,6 +49,7 @@ class IFDSUninitializedVariablesTest : public ::testing::Test {
   void TearDown() override {
     delete IRDB;
     delete TH;
+    delete PT;
     delete ICFG;
     delete UninitProblem;
```
vulder, line 54: The same could be applied here, regarding `unique_ptr` instead of `new` (see bellow).
Would be nice if you could quickly refactored the tests here and replace all the `new` calls with `unique_ptrs`, so we can prevent this kind of bug in the future. :+1:

vulder (changes requested):
Config file to make the SRC/BUILD path specific sounds like a good idea :+1: 
Furthermore, thanks for finding the leaks and fixing them.

unittests/TestUtils/TestConfig.h
benquike: Updated

unittests/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IFDSUninitializedVariablesTest.cpp
benquike: Updated. 

All the places using raw pointers are updated using unique_ptr.

vulder (approved):
Very nice, big thanks for cleaning that up :+1:

pdschubert (approved):
Many thanks! Looks good.

PR #226, opened 2020-12-21
Fixes config file generation location

vulder:
The generated config file needs to placed in the CURRENT_BINARY_DIR not
just the BINARY_DIR. In cases where phasar is embedded, e.g., in the case of an llvm in-tree build, the binary dirs differ and break the build.


pdschubert (approved):
Thanks.

PR #227, opened 2021-01-11
Improvements and minor changes in the shell scripts

vaibhav-hexa:
Hello,
This pull request includes few improvements and minor changes in the shell `scripts`. Please see the below list for the overview of the changes:

- Boost version check for installations with versions like 1._xy_.**0**
- Typo with `DO_UNIT_TESTS` corrected to `DO_UNIT_TEST`
- Added `|| exit ` for `cd` incase if it fails, so the script will exit
- Double-quoted the variables: `${var}` to `"${var}"`
- Command substitution with `$(...)` instead of `` `...` ``

Related issue: https://github.com/secure-software-engineering/phasar/issues/228


utils/init-submodules-release.sh
```diff
@@ -1,14 +1,14 @@
 #!/bin/bash
```
vulder, line 2:
```suggestion
function save_cd {
  cd "$@" || exit
}

```
I would put the exit check of the cd into a small function so its a) easier to write the code below without having to spell out exit and b) one cannot forget to call exit after a cd call.

If you want to make this function and future other util functions usable in all of the scripts define one util script for phasar that declares the functions and `sources` it into the other scripts.

utils/init-submodules-release.sh
```diff
@@ -1,14 +1,14 @@
...
 
 if git submodule status 2>&1 | grep -iq "fatal: Not a git repository (or any of the parent directories): .git"; then
-cd $(dirname $0)/../external/
+cd "$(dirname "$0")"/../external/ || exit
```
vulder, line 4:
```suggestion
save_cd "$(dirname "$0")"/../external/
```

utils/init-submodules-release.sh
```diff
@@ -1,14 +1,14 @@
...
 git clone git@github.com:google/googletest.git
-cd googletest/
+cd googletest/ || exit
```
vulder, line 6:
```suggestion
save_cd googletest/
```

utils/init-submodules-release.sh
```diff
@@ -1,14 +1,14 @@
...
 git checkout release-1.8.0
-cd -
+cd - || exit
 git clone git@github.com:nlohmann/json.git
-cd json/
+cd json/ || exit
 git checkout v3.4.0
-cd -
+cd - || exit
```
vulder, line 12:
```suggestion
save_cd -
```

utils/init-submodules-release.sh
```diff
@@ -1,14 +1,14 @@
...
+cd "$(dirname "$0")"/../external/ || exit
 git clone git@github.com:google/googletest.git
-cd googletest/
+cd googletest/ || exit
 git checkout release-1.8.0
-cd -
+cd - || exit
 git clone git@github.com:nlohmann/json.git
-cd json/
+cd json/ || exit
```
vulder, line 10:
```suggestion
save_cd json/
```

utils/init-submodules-release.sh
```diff
@@ -1,14 +1,14 @@
...
 
 if git submodule status 2>&1 | grep -iq "fatal: Not a git repository (or any of the parent directories): .git"; then
-cd $(dirname $0)/../external/
+cd "$(dirname "$0")"/../external/ || exit
 git clone git@github.com:google/googletest.git
-cd googletest/
+cd googletest/ || exit
 git checkout release-1.8.0
-cd -
+cd - || exit
```
vulder, line 8:
```suggestion
save_cd -
```

utils/init-submodules-release.sh
vaibhav-hexa: Thank you for the suggestions. I have created a new script as `./utils/safeCommandsSet.sh` and sourced it into other scripts which have `cd` calls. Other than the `cd` , I do not see another command which can be converted into `function`. 
In the future, other sets of `commands` can be replaced with new functions in the script.

utils/safeCommandsSet.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+function save_cd {
```
vulder, line 3:
```suggestion
function safe_cd {
```

I'm sorry, I introduce a typo here when suggesting the function name :see_no_evil:. It should be `safe_cd`.

vulder (changes requested):

utils/init-submodules-release.sh
vulder: Thanks looks good :+1:

utils/safeCommandsSet.sh
vaibhav-hexa: Ahh, got it, thanks!

vulder (approved):
LGTM :smiley:

vulder:
@pdschubert If you don't see any problems we could merge this

bootstrap.sh
```diff
@@ -96,48 +98,51 @@ else
...
+            boostlibnames=("libboost-system" "libboost-filesystem"
                     "libboost-graph" "libboost-program-options"
                     "libboost-log" "libboost-thread")
             additional_boost_libs=()
             for boost_lib in ${boostlibnames[@]}; do
-                dpkg -s "$boost_lib${DESIRED_BOOST_VERSION}" >/dev/null 2>&1 || additional_boost_libs+=("$boost_lib${DESIRED_BOOST_VERSION}")
+                dpkg -s "$boost_lib${DESIRED_BOOST_VERSION}" >/dev/null 2>&1 ||
+                dpkg -s "$boost_lib${DESIRED_BOOST_VERSION}.0" >/dev/null 2>&1 ||
+                additional_boost_libs+=("$boost_lib${DESIRED_BOOST_VERSION}") ||
+                additional_boost_libs+=("$boost_lib${DESIRED_BOOST_VERSION}.0")
```
pdschubert, line 109: Why do we need to explicitly add a `.0` at the end here?

pdschubert (changes requested):
Thanks for your improvements. I just found one thing that someone needs to explain to me. See my comment below.

bootstrap.sh
vaibhav-hexa: Hello @pdschubert ,  
Thank you for reviewing the changes. 

If an installed version of the boost library version is `1.73.0` then the current check fails to detect as it compares 1.73 with 1.73`.0` and the script exits. Considering the boost releases on https://www.boost.org/users/history/ , the releases are in a pattern of 1.xy.0. So, adding a `.0` check works fine.

pdschubert (approved):

PR #233, opened 2021-02-12
move call-to-return flow out of iteration of callee targets

MMory:


pdschubert (approved):
Wow, you're right. Thanks for pointing out that flaw. This may originated from a previous merge, but now it's fixed.

PR #235, opened 2021-02-16
Global Constructors and Destructors in ICFG

alexmeinhold:
Global constructors and destructors are stored in the global variables @llvm.global_ctors/@llvm.global_dtors in LLVM IR.
This pull request allows passing a parameter to the LLVMBasedICFG constructor to parse these global ctors/dtors and include them in the call graph so that they can be used for further analysis.


pdschubert:
I will introduce artificial control-flow edges such that we are able to correctly model global variables.

@MMory, I'll let you know it's done and you can review the code.

PR #239, opened 2021-03-09
Link against libLLVM.so

J03D03:
### Description
Switch from one library for each LLVM component to a single library for all components as recommended by LLVM.
> BUILD_SHARED_LIBS is only recommended for use by LLVM developers. If you want to build LLVM as a shared library, you should use the LLVM_BUILD_LLVM_DYLIB option.
https://llvm.org/docs/CMake.html

BUILD_SHARED_LIBS:DEFAULT=Off
LLVM_LINK_LLVM_DYLIB=ON ⇒ LLVM_BUILD_LLVM_DYLIB=ON

---

This PR also removes one extinct flag
```
CMake Warning:
  Manually-specified variables were not used by the project:

    LLVM_ENABLE_CXX1Y
```

### DIFF
--- a/utils/install-llvm.sh
+++ b/utils/install-llvm.sh
@@ -50,7 +50,7 @@ safe_cd "${build_dir}"/llvm-project/
 git checkout "${llvm_release}"
 mkdir -p build
 safe_cd build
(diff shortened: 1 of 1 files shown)


vulder (approved):

MMory (approved):
Thanks!

pdschubert (approved):
Great, thanks!

PR #301, opened 2021-04-14
Fix memory leak

blipper:


MMory:
Hi @blipper , thank you very much for your contributions. Is it feasible for you with little effort to raise your pull requests against the development branch? The master branch is considered the stable release branch, while the development branch receives all the development and fixes. Of course we want your contributions also in development. I just tried to alter this PR to development, but then there is a gazillion of changes which doesn't seem right. If that would be too much of work for you then I suggest we make a clone branch of master and then do cherry-picking into development. Maybe that would be the best way to go. What do you think?

blipper:
Hi

No problem to add future fixes to dev branch.  I am used to Google's model which is mostly branchless 😂

For various reasons I raised these prs just directly via the GitHub online editor.  I don't have a super easy way to raise them again though maybe the GitHub desktop client can help.

One approach is merge these changes into master and then rebase the current dev on the master and resolve the conflicts iteratively.   The vast majority are pretty innocuous so should be ok in master and any conflicts should be minor.

blipper:
Just want to follow up here so that I can get my fixes into Phasar?

pdschubert (approved):

PR #310, opened 2021-04-15
Add OpenMP task parallelism to LLVMPointsToSet global evaluation

blipper:
Simple implementation that locks the PointsToSets 
Binary needs to be linked with OpenMP 
See https://stackoverflow.com/questions/33357029/using-openmp-with-clang and https://stackoverflow.com/questions/10481623/how-to-tell-if-openmp-is-working

Gives 4-6x times speedup.  

Depends on changes in #309


blipper:
Ignore this for now.  Doesn't work.

PR #341, opened 2021-05-05
Only add TSAN for testing when compatible

vulder:
Only add TSAN when compiled with clang and ASAN is not active.

### DIFF
--- a/unittests/PhasarLLVM/DataFlowSolver/IfdsIde/CMakeLists.txt
+++ b/unittests/PhasarLLVM/DataFlowSolver/IfdsIde/CMakeLists.txt
@@ -12,9 +12,13 @@ set(ThreadedIfdsIdeSources
   EdgeFunctionSingletonFactoryTest.cpp
 )
 
-if(UNIX)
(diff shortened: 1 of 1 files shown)


pdschubert (changes requested):
Thanks! Besides checking the `LLVM_USE_SANITIZER` variable, could you please also check the presence of the address sanitizer flag in the `CMAKE_CXX_FLAGS` variable?
I'm currently trying to improve code quality and there's something lurking on the `f-ImproveOnClangTidySupport` branch that would lead to the same issue.

vulder:
I can do that if you like but I guess it would be better to introduce a PHASAR_USE_SANITIZER if you want sanitizer handling in phasar, so phasar can have one point of mapping sanitizer requests to the appropriate flags. For example, some sanitizer can/should have extra flags added to produce better results or want a blacklist file (to hide some ubsan errors that are falls positives or cannot be fixed by phasar).

Then we can check here also that PHASAR_USE_SANITIZER does not contain asan.

pdschubert:
That makes sense, I guess. We'll integrate something like that on the `f-ImproveOnClangTidySupport` :wink:

PR #347, opened 2021-05-18
Fixes Dockerfile

jplehr:
This PR adds a missing COPY to the Dockerfile to be (again) able to build the container.


pdschubert:
Thanks.

PR #348, opened 2021-05-28
Adds a query method for possible allocation sites to PointsToInfo

vulder:
Extends the PointsToInfo interface with a query method
`isInReachableAllocationSites`. This enables users to simply query if a
value is in the set of reachable allocation sites without needing to
return a copy of the set.


pdschubert (approved):
Thanks, Flo!

PR #392, opened 2021-07-24
Fix SEGV on due to nullptr reference when not passing LLVMPointsToInfo to LLVMBasedICFG parameter.

blipper:


blipper:
Wrong commit

PR #401, opened 2021-07-26
Enable CMAKE_CXX_CLANG_TIDY for lib and unitests

blipper:


blipper:
Is there any reason not to merge this?  I believe it will just generate warnings.

MMory:
Hi Martin,

we are in the process of supporting clang tidy, but we also take care of the warnings it reports. This is happening in a different branch, so I am closing this one here.

Cheers
Martin

PR #406, opened 2021-07-28
Fixes handling for non existent return values

vulder:
In some cases we don't get a return value, hence, we need to correctly
kill the flow facts at this point and not return an LCAFF.


pdschubert (approved):

pdschubert:
Many thanks!

Issue #411, opened 2021-08-05
Add save/load methods for LLVMPointsToSet


ga0:
**Is your feature request related to a problem? Please describe.**
LLVMPointsToSet takes too much time.

**Describe the solution you'd like**
Since parallelize LLVMPointsToSet is too hard  ( #305 ), providing methods to serialize/deserialize or save/load LLVMPointsToSet may be very useful.

```c++
    ProjectIRDB IRDB({path});
    LLVMPointsToSet PTS(IRDB, false, PointerAnalysisType::CFLAnders);
    std::ofstream PointsToSetCacheFile("/tmp/points_to_set");
    PTS.save(PointsToSetCacheFile);
```

Next time when we want to analyze the same project.

```c++
    ProjectIRDB IRDB({path});
    LLVMPointsToSet PTS(IRDB);
    std::ifstream PointsToSetCacheFile("/tmp/points_to_set");
    PTS.load(PointsToSetCacheFile);
```

This will save much time.

The PointsToSetMap is an unordered_map of llvm::Value* pointers. 
I don't known how to serializing these pointers' relations.
Can you give me some instructions?

fabianbs96:
Updates are in #441

PR #412, opened 2021-08-10
Add save/load for LLVMPointsToSet

ga0:
As mentioned in #411 , save/load methods for LLVMPointsToSet will be very useful, especially when processing large programs. We can load the saved data and no need to generate LLVMPointsToSet again.

### DIFF
--- a/lib/PhasarLLVM/Pointer/LLVMPointsToSet.cpp
+++ b/lib/PhasarLLVM/Pointer/LLVMPointsToSet.cpp
@@ -9,6 +9,7 @@
 
 #include <cassert>
 #include <iostream>
+#include <fstream>
 #include <memory>
 #include <type_traits>
 #include <unordered_set>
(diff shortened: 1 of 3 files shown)


unittests/PhasarLLVM/Pointer/LLVMPointsToSetTest.cpp
```diff
@@ -58,6 +58,25 @@ TEST(LLVMPointsToSet, Global_01) {
   std::cout << '\n';
 }
 
+TEST(LLVMPointsToSet, SaveAndLoad) {
+  std::vector<std::string> testcases = {"pointers/basic_01_cpp.ll", "pointers/global_01_cpp.ll"};
```
vulder, line 62: does this case test all the different cases that can happen during loading/storing?

lib/PhasarLLVM/Pointer/LLVMPointsToSet.cpp
```diff
@@ -59,6 +83,95 @@ LLVMPointsToSet::LLVMPointsToSet(ProjectIRDB &IRDB, bool UseLazyEvaluation,
...
+    ++NextID;
+  });
+
+  std::set<std::shared_ptr<std::unordered_set<const llvm::Value *>>> Printed;
+
+  OS << "[AnalyzedFunctions]" << std::endl;
+  for (auto &F : AnalyzedFunctions) {
+    OS << ValueToIdMap[F] << " ";
+  }
+  OS << std::endl;
```
vulder, line 114:
```suggestion
  OS << '\n';
```
I guess, we don't need a flush here.

lib/PhasarLLVM/Pointer/LLVMPointsToSet.cpp
```diff
@@ -59,6 +83,95 @@ LLVMPointsToSet::LLVMPointsToSet(ProjectIRDB &IRDB, bool UseLazyEvaluation,
...
+      // Print each set only once
+      continue;
+    }
+
+    Printed.insert(S);
+    for (const auto &V : *S) {
+      OS << ValueToIdMap[V] << " ";
+    }
+
+    OS << std::endl;
```
vulder, line 128:
```suggestion
    OS << '\n';
```
I guess, we don't need a flush here.

lib/PhasarLLVM/Pointer/LLVMPointsToSet.cpp
```diff
@@ -59,6 +83,95 @@ LLVMPointsToSet::LLVMPointsToSet(ProjectIRDB &IRDB, bool UseLazyEvaluation,
   }
 }
 
+LLVMPointsToSet::LLVMPointsToSet(ProjectIRDB &IRDB,
+                                 const std::string& PointsToSetFile)
+{
+  load(PointsToSetFile, IRDB);
+}
+
+void LLVMPointsToSet::save(const std::string &PointsToSetFile, ProjectIRDB &IRDB) {
```
vulder, line 92: Is there a particular reason why we encode the save/load and stream handling our selfs?
From my point of view, it would be easier to encode the data structure into llvms yaml traits and let them do the heavy lifting.

vulder (changes requested):

pdschubert:
Thanks for your contribution. This feature is now fully implemented in #441 and will soon be merged.

PR #416, opened 2021-08-25
Default IncludeGlobals to on for soundness

MMory:
With IncludeGlobals == false the call graph and points-to information may be unsound.


fabianbs96 (commented):
Yes; this is definitely true.  However, in order to make fully use of it, we'd need to change the entrypoint of every analysis problem to the default of `IRDB->getFunction(LLVMBasedICFG::GlobalCRuntimeModelName)`. Otherwise, the analysis would still ignore the global ctors/dtors. See [LLVMBasedICFGGlobCtorDtorTest](https://github.com/secure-software-engineering/phasar/blob/e895f42fce11d22e1aa2b95bde734e9c3d4878ac/unittests/PhasarLLVM/ControlFlow/LLVMBasedICFGGlobCtorDtorTest.cpp#L257) as a reference. We might want to create a shortcut for this

PR #419, opened 2021-08-25
Global calls need to be collected/constructed before callgraph is constructed

MMory:


fabianbs96 (commented):
As far as I can see, the call graph is getting constructed in the fixpoint loop below (Lines 180ff).
The `makeResolver` function only initializes the resolver object doing nothing but initializing the internal fields (IRDB, TH, ICF, PT). So this code movement should have no visible effect.

pdschubert (approved):

PR #425, opened 2021-10-19
Fixes linking issue, where PTS requires getNumGlobals

vulder:
This moves getNumGlobals into the ProjectIRDB header because the
current implementation of the PointsToSet constructor uses it, however,
phasar_db(ProjectIRDB) links against phasar_pointer, so we cannot link
PointsToSet against phasar_db. By defining the function in the header we
circumvent this issue, however, we should be careful to not introduce
cyclic dependencies in the future.


fabianbs96:
Hi, as far as I can see, the phasasr_db does not use phasar_pointer (as well as phasar_typehierarchy) at all and on my system I could remove them without any error. So I would suggest to delete these dependencies

PR #434, opened 2021-11-02
Fixes linking issues of taintconfig + json workaround

vulder:
Fixes linking issues of taintconfig and adds a workaround to include the
schema validator in LLVM_EXPORT for in tree builds, so llvm correctly
installs the schema lib alongside it's tools/libs.


MMory (approved):
made another commit to add the newly added deps to the respective .cmake files

pdschubert (approved):

PR #444, opened 2021-12-23
Enable clang-tidy support

MMory:


CMakeLists.txt
```diff
@@ -20,11 +20,17 @@ include(GNUInstallDirs)
 set_property(GLOBAL PROPERTY TARGET_MESSAGES OFF)
 
 if (NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Build mode ('Debug' or 'Release', default is 'Debug')" FORCE)
+  set(CMAKE_BUILD_TYPE "DebugSan" CACHE STRING "Build mode ('DebugSan' or 'Debug' or 'Release', default is 'DebugSan')" FORCE)
```
vulder, line 23: I'm not sure if debug + sanitizer is the best default. Compiling with sanitizers takes longer and slows down development. I would assume that most users use dev builds by default and only enable sanitizer if needed, at least that's what I do most of the time.

However, having a dev+san build always running in the CI is a good idea and should catch bugs before they are integrated into `development`.

CMakeLists.txt
```diff
@@ -237,8 +243,17 @@ endif()
 add_subdirectory(external/WALi-OpenNWA)
 include_directories(external/WALi-OpenNWA/Source/wali/include)
 
+# Set up clang-tidy to run during PhASAR's compilation to indicate code smells
+if (PHASAR_ENABLE_CLANG_TIDY_DURING_BUILD)
```
vulder, line 247: Just as food for thought, in our project we don't run tidy all the time during normal compilation but have a specific cmake target that runs tidy over the code base (`ninja tidy-foo`). I found this a bit more convenient as tidy can sometimes take quite a bit of time to complete.

vulder (approved):

CMakeLists.txt
pdschubert: That's a good point actually. I'll change that slightly to have plain `Debug` as default.

pdschubert (approved):

PR #461, opened 2021-12-23
clang-tidy: Fix Mono

MMory:
~~please merge after #460~~

PR #462, opened 2021-12-23
clang-tidy: Fix WPDS

MMory:
~~please merge after #461~~


vulder (commented):
I guess you want to keep the d1, d2, d3s lower case?
If yes, we should add an exception for clang-tidy instead of putting NOLINT everywhere

pdschubert (approved):

PR #469, opened 2021-12-24
clang-tidy: Fix PhasarConfig

MMory:


include/phasar/Config/Configuration.h
```diff
@@ -35,51 +35,52 @@ namespace psr {
 class PhasarConfig {
 public:
   /// Current Phasar version
-  static std::string PhasarVersion() { return XSTR(PHASAR_VERSION); }
+  static std::string PhasarVersion() { return XSTR(PHASAR_VERSION); } // NOLINT
```
vulder, line 38: What's the reason we need no lint's for these functions?

PR #470, opened 2021-12-31
Fix for unescaped dot-labels

ste-lam:
Fixes the problem when an edge or vertex-label contains special characters


pdschubert (approved):
Thanks for the fix.

PR #471, opened 2022-01-05
LCA improvements

fabianbs96:
Several updates in the IDELinearConstantAnalysis


pdschubert:
Thanks for the improvements. However, the unit tests cannot be build since a required `operator==` is missing:
```
-c ../unittests/PhasarLLVM/ControlFlow/LLVMBasedICFGGlobCtorDtorTest.cpp
In file included from ../unittests/PhasarLLVM/ControlFlow/LLVMBasedICFGGlobCtorDtorTest.cpp:10:
../external/googletest/googletest/include/gtest/gtest.h:1545:11: error: invalid operands to binary expression ('const int' and 'const psr::LatticeDomain<long>')
  if (lhs == rhs) {
```

pdschubert:
I'm working on a fix.

pdschubert (approved):

PR #474, opened 2022-01-10
clang-tidy Fix IDE solver

MMory:


include/phasar/PhasarLLVM/Pointer/LLVMPointsToGraph.h
```diff
@@ -76,7 +76,7 @@ class LLVMPointsToGraph : public LLVMPointsToInfo {
     /// This may contain a call or invoke instruction.
     const llvm::Value *V = nullptr;
     EdgeProperties() = default;
-    EdgeProperties(const llvm::Value *V);
+    EdgeProperties(const llvm::Value *v);
```
vulder, line 79: Why did the name change from upper to lower case? Should not uppercase be the correct llvm style?

lib/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDESecureHeapPropagation.cpp
```diff
@@ -305,9 +314,9 @@ IDESecureHeapPropagation::SHPGenEdgeFn::getInstance(l_t Val) {
   static std::array<std::shared_ptr<SHPGenEdgeFn>, 3> Cache = {nullptr, nullptr,
                                                                nullptr};
   auto Ind = static_cast<std::underlying_type<l_t>::type>(Val);
-  if (!Cache[Ind]) {
-    Cache[Ind] = std::make_shared<SHPGenEdgeFn>(Val);
+  if (!Cache.at(Ind)) {
+    Cache.at(Ind) = std::make_shared<SHPGenEdgeFn>(Val);
```
vulder, line 318:
```suggestion
    Cache[Ind] = std::make_shared<SHPGenEdgeFn>(Val);
```
If we already did a bounce check in the previous line, this one should not be necessary.

vulder (changes requested):

PR #478, opened 2022-01-10
clang-tidy: fixes for remaining files

MMory:


lib/Controller/AnalysisController.cpp
```diff
@@ -130,17 +130,18 @@ void AnalysisController::executeVariational() {}
...
-    if (std::holds_alternative<DataFlowAnalysisType>(_DataFlowAnalysis)) {
-      auto getTaintConfig = [&]() {
-        if (!AnalysisConfigPath.empty()) {
-          return TaintConfig(IRDB, parseTaintConfig(AnalysisConfigPath));
-        }
-        return TaintConfig(IRDB);
-      };
-      auto DataFlowAnalysis = std::get<DataFlowAnalysisType>(_DataFlowAnalysis);
+    if (std::holds_alternative<DataFlowAnalysisType>(DFA)) {
+      auto getTaintConfig // NOLINT
```
vulder, line 137: What's the problem here?

pdschubert: Mhm, do you have a good idea here? The issue is that we store a lambda to a variable which triggers the invalid case styling rule since `getTaintConfig` is actually a variable rather than an ordinary function.

vulder: So using `GetTaintConfig` as variable name would solve this?

pdschubert (approved):

PR #483, opened 2022-01-17
Fixed Dockerfile 

janniclas:
Updated LLVM version in the Dockerfile to match the project's requirements.
Also changed the ubuntu image version from latest to a fixed version (20.04), since the newest ubuntu version 21.10 doesn't work with LLVM 12.0.0 and they might update their latest tag at some point.

I'm currently working on a multistage Dockerfile to increase performance. I experimented with GitHub Actions and building the current Dockerfile with an action takes around 4 hours, which is not sufficient.


pdschubert (approved):

pdschubert:
Many thanks!

PR #490, opened 2022-02-15
Fix points to set

fabianbs96:
Fix a use-after-free error in the LLVMPointsToSet


pdschubert (commented):
That does not look too bad at first glance.

Could you make it a 'real' pull-request such that I can try it out on the example that triggered the bug?

pdschubert (approved):

PR #496, opened 2022-04-04
Add LLVM 14 support

MMory:


lib/PhasarLLVM/Utils/Annotation.cpp
```diff
@@ -55,7 +56,8 @@ uint64_t VarAnnotation::getLine() const {
 llvm::StringRef VarAnnotation::retrieveString(unsigned Idx) const {
   if (const auto *ConstExpr = llvm::dyn_cast<llvm::ConstantExpr>(
           AnnotationCall->getArgOperand(Idx))) {
-    if (ConstExpr->isGEPWithNoNotionalOverIndexing()) {
+    if (true // ConstExpr->isGEPWithNoNotionalOverIndexing()
```
fabianbs96, line 59:
```suggestion
    if (llvm::isa<llvm::GEPOperator>(ConstExpr)
```

lib/PhasarLLVM/Utils/Annotation.cpp
```diff
@@ -109,7 +111,8 @@ llvm::StringRef GlobalAnnotation::retrieveString(unsigned Idx) const {
   const auto *AnnotationGepOp = AnnotationStruct->getOperand(Idx);
   if (const auto *ConstExpr =
           llvm::dyn_cast<llvm::ConstantExpr>(AnnotationGepOp)) {
-    if (ConstExpr->isGEPWithNoNotionalOverIndexing()) {
+    if (true // ConstExpr->isGEPWithNoNotionalOverIndexing()
```
fabianbs96, line 114: Same as above

lib/PhasarLLVM/DataFlowSolver/Mono/Problems/InterMonoFullConstantPropagation.cpp
```diff
@@ -79,13 +79,13 @@ InterMonoFullConstantPropagation::callFlow(
       llvm::isa<llvm::InvokeInst>(CallSite)) {
     const auto *CS = llvm::cast<llvm::CallBase>(CallSite);
     // early exit; varargs not handled yet
-    if (CS->getNumArgOperands() == 0 || Callee->isVarArg()) {
+    if (CS->arg_size() == 0 || Callee->isVarArg()) {
```
fabianbs96, line 82:
```suggestion
    if (CS->arg_empty() || Callee->isVarArg()) {
```

lib/PhasarLLVM/Utils/Annotation.cpp
MMory: Is that equivalent? What is the purpose of this if anyhow?

fabianbs96: No, it is not equivalent, but as close as we can get. 
When you pass a constant string to a function, the string data is a global variable of an array type, so you always pass a constant GEP to that global instead. 

In the ctor of VarAnnotation we don't check that the AnnotationCall is indeed a call to an llvm.var.annotation intrinsic, so in theory this class could be reused for some other purpose. However, in our use-case, this if-check should always succeed

vulder (approved):
Except for the `isGEPWithNoNotionalOverIndexing` issue, this looks fine to me.

fabianbs96 (approved):

PR #499, opened 2022-05-06
fix VTables for collapsed types. 

MMory:
llvm-link collapsed multiple C++  types with identical members to one single LLVM type. The current  virtual function call resolution then only considers the VTable of the LLVM type that represents all of the collapsed type. Now, we look for all VTables that the function pointer aliases with and extract the possible callees from them.

### DIFF
--- a/include/phasar/PhasarLLVM/ControlFlow/Resolver/Resolver.h
+++ b/include/phasar/PhasarLLVM/ControlFlow/Resolver/Resolver.h
@@ -29,11 +29,15 @@ class Function;
 class StructType;
 } // namespace llvm
 
+namespace std {
+template <class T> class optional;
+}
+
 namespace psr {
 class ProjectIRDB;
 class LLVMTypeHierarchy;
 
-int getVFTIndex(const llvm::CallBase *CallSite);
+std::optional<unsigned> getVFTIndex(const llvm::CallBase *CallSite);
 
 const llvm::StructType *getReceiverType(const llvm::CallBase *CallSite);
 
--- a/include/phasar/PhasarLLVM/ControlFlow/Resolver/Resolver.h
+++ b/include/phasar/PhasarLLVM/ControlFlow/Resolver/Resolver.h
@@ -17,6 +17,7 @@
 #ifndef PHASAR_PHASARLLVM_CONTROLFLOW_RESOLVER_RESOLVER_H_
 #define PHASAR_PHASARLLVM_CONTROLFLOW_RESOLVER_RESOLVER_H_
 
+#include <optional>
 #include <set>
 #include <string>
 
@@ -28,10 +29,6 @@ class CallBase;
 class Function;
 class StructType;
 } // namespace llvm
-
-namespace std {
-template <class T> class optional;
-}
 
 namespace psr {
 class ProjectIRDB;
--- a/include/phasar/PhasarLLVM/TypeHierarchy/LLVMTypeHierarchy.h
+++ b/include/phasar/PhasarLLVM/TypeHierarchy/LLVMTypeHierarchy.h
@@ -88,19 +88,16 @@ public:
   using out_edge_iterator = boost::graph_traits<bidigraph_t>::out_edge_iterator;
   using in_edge_iterator = boost::graph_traits<bidigraph_t>::in_edge_iterator;
 
-  static const std::string StructPrefix;
-
-  static const std::string ClassPrefix;
-
-  static const std::string VTablePrefix;
-
-  static const std::string VTablePrefixDemang;
-
-  static const std::string TypeInfoPrefix;
-
-  static const std::string TypeInfoPrefixDemang;
-
-  static const std::string PureVirtualCallName;
+  static inline constexpr llvm::StringLiteral StructPrefix = "struct.";
+  static inline constexpr llvm::StringLiteral ClassPrefix = "class.";
+  static inline constexpr llvm::StringLiteral VTablePrefix = "_ZTV";
+  static inline constexpr llvm::StringLiteral VTablePrefixDemang =
+      "vtable for ";
+  static inline constexpr llvm::StringLiteral TypeInfoPrefix = "_ZTI";
+  static inline constexpr llvm::StringLiteral TypeInfoPrefixDemang =
+      "typeinfo for ";
+  static inline constexpr llvm::StringLiteral PureVirtualCallName =
+      "__cxa_pure_virtual";
 
 private:
   bidigraph_t TypeGraph;
--- a/include/phasar/PhasarLLVM/TypeHierarchy/LLVMTypeHierarchy.h
+++ b/include/phasar/PhasarLLVM/TypeHierarchy/LLVMTypeHierarchy.h
@@ -100,6 +100,8 @@ public:
 
   static const std::string TypeInfoPrefixDemang;
 
+  static const std::string PureVirtualCallName;
+
 private:
   bidigraph_t TypeGraph;
   std::unordered_map<const llvm::StructType *, vertex_t> TypeVertexMap;
--- a/include/phasar/PhasarLLVM/TypeHierarchy/LLVMTypeHierarchy.h
+++ b/include/phasar/PhasarLLVM/TypeHierarchy/LLVMTypeHierarchy.h
@@ -88,6 +88,18 @@ public:
   using out_edge_iterator = boost::graph_traits<bidigraph_t>::out_edge_iterator;
   using in_edge_iterator = boost::graph_traits<bidigraph_t>::in_edge_iterator;
 
+  static const std::string StructPrefix;
+
+  static const std::string ClassPrefix;
+
+  static const std::string VTablePrefix;
+
+  static const std::string VTablePrefixDemang;
+
+  static const std::string TypeInfoPrefix;
+
+  static const std::string TypeInfoPrefixDemang;
+
 private:
   bidigraph_t TypeGraph;
   std::unordered_map<const llvm::StructType *, vertex_t> TypeVertexMap;
@@ -101,18 +113,6 @@ private:
   std::unordered_map<std::string, const llvm::GlobalVariable *> ClearNameTIMap;
   // map from clearname to vtable variable
   std::unordered_map<std::string, const llvm::GlobalVariable *> ClearNameTVMap;
-
-  static const std::string StructPrefix;
-
-  static const std::string ClassPrefix;
-
-  static const std::string VTablePrefix;
-
-  static const std::string VTablePrefixDemang;
-
-  static const std::string TypeInfoPrefix;
-
-  static const std::string TypeInfoPrefixDemang;
 
   static std::string removeStructOrClassPrefix(const llvm::StructType &T);
 
--- a/include/phasar/PhasarLLVM/TypeHierarchy/LLVMVFTable.h
+++ b/include/phasar/PhasarLLVM/TypeHierarchy/LLVMVFTable.h
@@ -87,7 +87,7 @@ public:
   };
 
   [[nodiscard]] static std::vector<const llvm::Function *>
-  getVFVectorFromIRVTable(const llvm::ConstantStruct *);
+  getVFVectorFromIRVTable(const llvm::ConstantStruct &);
 };
 
 } // namespace psr
--- a/include/phasar/PhasarLLVM/TypeHierarchy/LLVMVFTable.h
+++ b/include/phasar/PhasarLLVM/TypeHierarchy/LLVMVFTable.h
@@ -19,6 +19,7 @@
 
 namespace llvm {
 class Function;
+class ConstantStruct;
 } // namespace llvm
 
 namespace psr {
@@ -84,6 +85,9 @@ public:
   end() const {
     return VFT.end();
   };
+
+  [[nodiscard]] static std::vector<const llvm::Function *>
+  getVFVectorFromIRVTable(const llvm::ConstantStruct *);
 };
 
 } // namespace psr
--- a/lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
+++ b/lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
@@ -90,8 +90,8 @@ auto OTFResolver::resolveVirtualCall(const llvm::CallBase *CallSite)
   LOG_IF_ENABLE(BOOST_LOG_SEV(lg::get(), DEBUG)
                 << "Call virtual function: " << llvmIRToString(CallSite));
 
-  auto VtableIndex = getVFTIndex(CallSite);
-  if (VtableIndex < 0) {
+  auto RetrievedVtableIndex = getVFTIndex(CallSite);
+  if (!RetrievedVtableIndex.has_value()) {
     // An error occured
     LOG_IF_ENABLE(BOOST_LOG_SEV(lg::get(), DEBUG)
                   << "Error with resolveVirtualCall : impossible to retrieve "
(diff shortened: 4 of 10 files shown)


include/phasar/PhasarLLVM/TypeHierarchy/LLVMVFTable.h
```diff
@@ -84,6 +85,9 @@ class LLVMVFTable : public VFTable<const llvm::Function *> {
   end() const {
     return VFT.end();
   };
+
+  [[nodiscard]] static std::vector<const llvm::Function *>
+  getVFVectorFromIRVTable(const llvm::ConstantStruct *);
```
vulder, line 90:
```suggestion
  [[nodiscard]] static std::vector<const llvm::Function *>
  getVFVectorFromIRVTable(const llvm::ConstantStruct &);
```
The impl does not work with a nullptr and also does not check for it, so making this clear through the API by passing a reference would be a good idea.

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
```diff
@@ -48,4 +50,33 @@ nlohmann::json LLVMVFTable::getAsJson() const {
...
+          if (auto *BC = llvm::dyn_cast<llvm::BitCastOperator>(CE)) {
+            // if the entry is a GlobalAlias, get its Aliasee
+            auto *ENTRY = BC->getOperand(0);
+            while (auto *GA = llvm::dyn_cast<llvm::GlobalAlias>(ENTRY)) {
+              ENTRY = GA->getAliasee();
+            }
+            auto *F = llvm::dyn_cast<llvm::Function>(ENTRY);
+            VFS.push_back(F);
+          } else {
+            VFS.push_back(nullptr);
```
vulder, line 71: What is the semantics of adding a `nullptr` here? That we did not find a function?

Callers that work on the result vector need to be aware that it may contain `nullptrs`.

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
```diff
@@ -48,4 +50,33 @@ nlohmann::json LLVMVFTable::getAsJson() const {
   return J;
 }
 
+std::vector<const llvm::Function *>
+LLVMVFTable::getVFVectorFromIRVTable(const llvm::ConstantStruct *VT) {
+  std::vector<const llvm::Function *> VFS;
+  for (const auto &Op : VT->operands()) {
+    if (auto *CA = llvm::dyn_cast<llvm::ConstantArray>(Op)) {
+      for (auto It = CA->operands().begin() + 2; It != CA->operands().end();
```
vulder, line 58:
```suggestion
      for (auto It = std::advance(CA->operands().begin(), 2); It != CA->operands().end();
```
+ This needs also a comment or constant name to make clear why we start not on the first but the third operand.

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
```diff
@@ -48,4 +50,33 @@ nlohmann::json LLVMVFTable::getAsJson() const {
...
+           ++It) {
+        auto &COp = *It;
+        if (auto *CE = llvm::dyn_cast<llvm::ConstantExpr>(COp)) {
+          if (auto *BC = llvm::dyn_cast<llvm::BitCastOperator>(CE)) {
+            // if the entry is a GlobalAlias, get its Aliasee
+            auto *ENTRY = BC->getOperand(0);
```
vulder, line 64: Why is `ENTRY` spellend in upper case letters? Is this an exception to the naming schema?

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
```diff
@@ -48,4 +50,33 @@ nlohmann::json LLVMVFTable::getAsJson() const {
...
+std::vector<const llvm::Function *>
+LLVMVFTable::getVFVectorFromIRVTable(const llvm::ConstantStruct *VT) {
+  std::vector<const llvm::Function *> VFS;
+  for (const auto &Op : VT->operands()) {
+    if (auto *CA = llvm::dyn_cast<llvm::ConstantArray>(Op)) {
+      for (auto It = CA->operands().begin() + 2; It != CA->operands().end();
+           ++It) {
+        auto &COp = *It;
+        if (auto *CE = llvm::dyn_cast<llvm::ConstantExpr>(COp)) {
+          if (auto *BC = llvm::dyn_cast<llvm::BitCastOperator>(CE)) {
```
vulder, line 62: As we only read from `COp`, could we mark it `const`?

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
```diff
@@ -105,29 +105,34 @@ auto OTFResolver::resolveVirtualCall(const llvm::CallBase *CallSite)
...
-      PossibleCallTargets.insert(Target);
+      auto PTS = PT.getPointsToSet(CallSite->getCalledOperand(), CallSite);
+      for (const auto *P : *PTS) {
+        if (auto *PGV = llvm::dyn_cast<llvm::GlobalVariable>(P)) {
+          if (PGV->hasName() && PGV->getName().startswith("_ZTV") &&
+              PGV->hasInitializer()) {
+            if (auto *PCS = llvm::dyn_cast<llvm::ConstantStruct>(
+                    PGV->getInitializer())) {
+              auto VFs = LLVMVFTable::getVFVectorFromIRVTable(PCS);
+              if (VtableIndex < 0 || VtableIndex >= VFs.size()) {
```
vulder, line 121: Is `VtableIndex < 0` an actual "wanted" condition?
If not, maybe we should put this check as an assert.

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
```diff
@@ -105,29 +105,34 @@ auto OTFResolver::resolveVirtualCall(const llvm::CallBase *CallSite)
...
 
-  for (const auto *PossibleTypeStruct : PossibleTypes) {
-    const auto *Target =
-        getNonPureVirtualVFTEntry(PossibleTypeStruct, VtableIndex, CallSite);
-    if (Target) {
-      PossibleCallTargets.insert(Target);
+      auto PTS = PT.getPointsToSet(CallSite->getCalledOperand(), CallSite);
+      for (const auto *P : *PTS) {
+        if (auto *PGV = llvm::dyn_cast<llvm::GlobalVariable>(P)) {
+          if (PGV->hasName() && PGV->getName().startswith("_ZTV") &&
```
vulder, line 116:
```suggestion
          if (PGV->hasName() && PGV->getName().startswith(LLVMTypeHierarchy::VTablePrefix) &&
```
This should use the prefix constant

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
```diff
@@ -105,29 +105,34 @@ auto OTFResolver::resolveVirtualCall(const llvm::CallBase *CallSite)
...
+              PGV->hasInitializer()) {
+            if (auto *PCS = llvm::dyn_cast<llvm::ConstantStruct>(
+                    PGV->getInitializer())) {
+              auto VFs = LLVMVFTable::getVFVectorFromIRVTable(PCS);
+              if (VtableIndex < 0 || VtableIndex >= VFs.size()) {
+                continue;
+              }
+              auto *Callee = VFs[VtableIndex];
+              if (Callee == nullptr || !Callee->hasName() ||
+                  Callee->getName() == "__cxa_pure_virtual") {
```
vulder, line 126: `` should probably also have it's own constant `PureVirtualCallName`.

I'm not sure how `phasar` handles this, but keep in mind that `__cxa_pure_virtual` is normally only used in the ItaniumABI not with Microsoft. I'm not sure how `llvm` implements this, i.e., if the they initially use `__cxa_pure_virtual` and later emit the Microsoft version or if the ABI difference is already visible at the IR level.

vulder (changes requested):

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
MMory: The semantics is that the index into the vtable at the call site is consistent with the vector of vtable entries that we are building here. Also at this time the callers are aware of the possible nullptrs.

include/phasar/PhasarLLVM/TypeHierarchy/LLVMVFTable.h
MMory: Yeah makes sense, I'll make that change not in github though.

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
MMory: What is the advantage of using std::advance? Bounds checking?
Index 0 is base offset, index 1 is RTTI, then the actual function entries start. But yeah, this should not just be a magic constant.

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
MMory: Yeah, this needs to be altered - this code fragment is code that existed previously and was refactored into this static function, see rest of diff. Nevertheless, the name is bad.

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
MMory: I think so, yes.

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
MMory: I am not sure about this, depends on code that is generated.

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
MMory: agreed.

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
MMory: The ABI difference is visible at IR level.

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
vulder: No, there is no bounds checking (or other performance penalty). The advantage of `advance` is that it's handing the iterator correctly based on it's iterator categorie and does not solely rely on pointer arithmetic. It makes the code a bit more readable. Plus, it's easier to handle for some static analyzers :wink:, as some are not super good in handling pointer arithmetic.

For me, it was a magic constant :)

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
vulder: What code would need/use a negative index into a VTable? This would be highly undefined behavior.

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
fabianbs96: Shouldn't it then be `std::next` ?. As far as I know, std::advance only increments in-place

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
fabianbs96: Actually, the current implementation returns negative indices as error codes, so we should check for that

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
MMory: And the +2 is not correct? It is also not pointer arithmetic, but operator+, isn't it? I really don't see the added value to use these functions and not just + 2.

fabianbs96: The +2 only works for rancom-access iterators, not in general as these STL functions do

vulder: @fabianbs96 Yes, I mixed them up in my head, it's `std::next` if you need the iterator returned.

@MMory No, `+2` is not always correct, everything depends on the iterator category. For example, where a RandomAccessIterator requires `operator+` to be implemented, Forward or Bidirectional iterators do only need to supply operator++ (and operator-- for the Bidirectional one). So, to not make mistakes or let the code break when underlying stuff get's changed/refactored underneath, using `std::next` or `std::advance` is advised.

Just to make this clear, this does not cost performance the std-library will pick the most optimal implementation for the specified iterator and the compiler will optimize everything, so in the end there should be no performance penalty.

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
vulder: Then, this sounds for a good place to add an assert to not run into trouble later.
We can still keep the check, so not to crash during a release build but I would assume we want to know if an error code is reaching this location in the code.

lib/PhasarLLVM/TypeHierarchy/LLVMVFTable.cpp
MMory: got it, changing it. Thanks, again something learned :)

lib/PhasarLLVM/ControlFlow/Resolver/OTFResolver.cpp
MMory: Actually it was really pain for me to see these negative error codes, terrible! I'm gonna do this right with std::optional now.

lib/PhasarLLVM/ControlFlow/Resolver/Resolver.cpp
```diff
@@ -30,23 +31,23 @@ using namespace psr;
...
+    return std::nullopt;
   }
   if (auto *CI = llvm::dyn_cast<llvm::ConstantInt>(GEP->getOperand(1))) {
     return CI->getZExtValue();
   }
-  return -3;
+  return std::nullopt;
 }
 
 const llvm::StructType *getReceiverType(const llvm::CallBase *CallSite) {
```
fabianbs96, line 53: This might not be relevant for this PR, but I noticed that `getReceiverType()` is wrong: The receiver (i.e. the `this` pointer) is not always the first parameter (index 0). In case of `sret` (return of large or non-trially movable types) the return value is the first parameter and the receiver is the second...

include/phasar/PhasarLLVM/ControlFlow/Resolver/Resolver.h
```diff
@@ -29,11 +29,15 @@ class Function;
 class StructType;
 } // namespace llvm
 
+namespace std {
+template <class T> class optional;
+}
+
```
vulder, line 35:
```suggestion
```

Please include optional

vulder (changes requested):

include/phasar/PhasarLLVM/TypeHierarchy/LLVMTypeHierarchy.h
```diff
@@ -88,6 +88,20 @@ class LLVMTypeHierarchy
...
+
+  static const std::string VTablePrefix;
+
+  static const std::string VTablePrefixDemang;
+
+  static const std::string TypeInfoPrefix;
+
+  static const std::string TypeInfoPrefixDemang;
+
+  static const std::string PureVirtualCallName;
```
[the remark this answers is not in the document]
pdschubert: Done.

include/phasar/PhasarLLVM/ControlFlow/Resolver/Resolver.h
pdschubert: Done.

pdschubert (approved):

PR #502, opened 2022-05-10
Fix getReceiverType() 

fabianbs96:
The receiver (i.e. the this pointer) is not always the first parameter (index 0). In case of sret (return of large or non-trially movable types) the return value is the first parameter and the receiver is the second


pdschubert (approved):
Thanks!

PR #503, opened 2022-05-12
New boost-less logger, getting rid of <iostream>

MMory:
This large PR introduces a new logger that uses llvm streams and gets rid of the dependency to boost::log and thereby yet another transitive dependency to boost::filesystem (with boost::dll remaining as the last one).

The new logger supports logging to stdout, stderr and file. It supports logging by level, by category or both. See Logging.h for the new macro interfaces. The closest matching and most recently set output stream is used.

In the course of this, removing std::cout and std::cerr from the code base as well as std::ostreams wherever this is possible, which, unfortunately, is not everywhere.

For reviewer's convenience, I have split the change kinds to various commits in this branch, which should make your life easier. It's many changes, but most of them are pretty obvious.


include/phasar/Utils/Logger.h
```diff
@@ -45,45 +43,131 @@ enum SeverityLevel {
...
+  static SeverityLevel getLoggerFilterLevel();
+
+  static bool isLoggingEnabled();
+
+  static void enable();
+
+  static void disable();
+
+  static llvm::raw_ostream &
+  getLogStream(const std::optional<SeverityLevel> &Level,
```
fabianbs96, line 62:
```suggestion
  getLogStream(std::optional<SeverityLevel> Level,
```

std::optional<SeverityLevel> is trivially copyable and small enough to be passed by value

include/phasar/Utils/Logger.h
```diff
@@ -45,45 +43,131 @@ enum SeverityLevel {
...
+               const std::optional<llvm::StringRef> &Category);
+
+  static bool logCategory(const llvm::StringRef &Category,
+                          const std::optional<SeverityLevel> &Level);
```
fabianbs96, line 66:
```suggestion
  static bool logCategory(llvm::StringRef Category,
                          std::optional<SeverityLevel> Level);
```

llvm::StringRef should be passed by value

include/phasar/Utils/Logger.h
```diff
@@ -45,45 +43,131 @@ enum SeverityLevel {
...
+
+  static void addLinePrefix(llvm::raw_ostream &,
+                            const std::optional<SeverityLevel> &Level,
```
fabianbs96, line 69: same as above

include/phasar/Utils/Logger.h
```diff
@@ -45,45 +43,131 @@ enum SeverityLevel {
...
+                            const std::optional<std::string> &Category);
+
+  static void initializeStdoutLogger(
+      const std::optional<SeverityLevel> &Level = std::nullopt,
```
fabianbs96, line 73: same

include/phasar/Utils/Logger.h
```diff
@@ -45,45 +43,131 @@ enum SeverityLevel {
...
+      const std::optional<std::string> &Category = std::nullopt);
+
+  static void initializeStderrLogger(
+      const std::optional<SeverityLevel> &Level = std::nullopt,
```
fabianbs96, line 77: same

include/phasar/Utils/Logger.h
```diff
@@ -45,45 +43,131 @@ enum SeverityLevel {
...
+      CategoriesToStreamVariant;
+  static inline std::map<std::optional<SeverityLevel>,
+                         std::variant<StdStream, std::string>>
+      LevelsToStreamVariant;
+  static inline SeverityLevel LogFilterLevel = DEBUG;
+  static inline llvm::StringMap<std::unique_ptr<llvm::raw_fd_ostream>>
+      LogfileStreams;
+  // static inline auto StartTime = std::chrono::steady_clock::now();
+  [[nodiscard]] static llvm::raw_ostream &
+  getLogStream(const std::optional<SeverityLevel> &Level,
```
fabianbs96, line 100: Same as above

include/phasar/Utils/Logger.h
```diff
@@ -45,45 +43,131 @@ enum SeverityLevel {
...
+               const std::map<std::optional<SeverityLevel>,
+                              std::variant<StdStream, std::string>>
+                   &PassedLevelsToStreamVariant);
+  [[nodiscard]] static llvm::raw_ostream &getLogStreamFromStreamVariant(
+      const std::variant<StdStream, std::string> &StreamVariant);
+};
 
-extern SeverityLevel LogFilterLevel;
+std::string toString(const SeverityLevel &Level);
```
fabianbs96, line 108:
```suggestion
std::string toString(SeverityLevel Level);
```

lib/Utils/Logger.cpp
```diff
@@ -51,74 +47,166 @@ std::string toString(const SeverityLevel &Level) {
...
+          Filename, EC, llvm::sys::fs::OpenFlags::OF_ChildInherit);
+    }
+    // Following
+    // https://stackoverflow.com/questions/41699343/how-do-i-test-that-an-stderror-code-is-not-an-error
+    if (EC) {
+      LogfileStreams[Filename] = nullptr;
+      llvm::errs() << "Failed to open logfile: " << Filename << '\n';
+      llvm::errs() << EC.message() << '\n';
+      return false;
+    }
```
fabianbs96, line 107:
```suggestion

  std::error_code EC;
  auto [It, Inserted] = [&] {
    if (Append) {
      return LogfileStreams.try_emplace(
          Filename, Filename, EC,
          llvm::sys::fs::OpenFlags::OF_Append |
              llvm::sys::fs::OpenFlags::OF_ChildInherit);
    }

    return LogfileStreams.try_emplace(
        Filename, Filename, EC, llvm::sys::fs::OpenFlags::OF_ChildInherit);
  }();

  if (!Inserted) {
    return true;
  }

  if (EC) {
    LogfileStreams.erase(Filename);
    llvm::errs() << "Failed to open logfile: " << Filename << '\n';
    llvm::errs() << EC.message() << '\n';
    return false;
  }
```

To compensate the missing `unique_ptr` indirection from the comment in the header

fabianbs96 (commented):
I am not done with the full review, just an intermediate state...

include/phasar/Utils/Logger.h
```diff
@@ -45,45 +43,131 @@ enum SeverityLevel {
...
 
 #ifdef DYNAMIC_LOG
-BOOST_LOG_INLINE_GLOBAL_LOGGER_DEFAULT(
-    lg, boost::log::sources::severity_logger<SeverityLevel>)
-// For performance reason, we want to disable any formatting computation
-// that would go straight into logs if logs are deactivated
-// This macro does just that
-#define LOG_IF_ENABLE_BOOL(condition, computation)                             \
+
```
vulder, line 111: Is the `DYNAMIC_LOG` options still necessary? What's the reasoning behind it?

lib/Utils/Logger.cpp
```diff
@@ -51,74 +47,166 @@ std::string toString(const SeverityLevel &Level) {
...
+llvm::raw_ostream &
+Logger::getLogStream(const std::optional<SeverityLevel> &Level,
+                     const std::map<std::optional<SeverityLevel>,
+                                    std::variant<StdStream, std::string>>
+                         &PassedLevelsToStreamVariant) {
+  if (Level.has_value()) {
+    std::optional<SeverityLevel> ClosestLevel = std::nullopt;
+    for (const auto &[LevelThreshold, _] : PassedLevelsToStreamVariant) {
+      if (LevelThreshold <= Level.value()) {
+        if (ClosestLevel == std::nullopt ||
```
vulder, line 134:
```suggestion
        if (ClosestLevel ||
```
The `bool`-ness check will automatically determine if the optional is empty.

lib/Utils/Logger.cpp
```diff
@@ -51,74 +47,166 @@ std::string toString(const SeverityLevel &Level) {
...
+    // }
+  if (Category.has_value()) {
+    OS << '[' << Category.value() << ']';
+  } // else {
+  //   OS << ' ';
+  // }
+  OS << ' ';
 }
 
+void initializeLogger(bool UseLogger, const std::string &LogFile) {}
```
vulder, line 206: Remove this one?

lib/Utils/Logger.cpp
```diff
@@ -51,74 +47,166 @@ std::string toString(const SeverityLevel &Level) {
...
+
+void Logger::enable() { LoggingEnabled = true; }
+
+void Logger::disable() { LoggingEnabled = true; }
```
vulder, line 210: These two could be moved into the header, as they are tiny.

vulder (changes requested):
It's a "small" PR for man but a giant leap for phasarkind :laughing: 

Very nice work, bit thx to you @MMory  :+1: 

PS: my `OS` remarks often affect the whole file, I was just to lay to mark every occurence

lib/Utils/Logger.cpp
fabianbs96: Yes, but please make sure not to forget the little `!`

include/phasar/Utils/Logger.h
fabianbs96: The original reason for introducint it was, that even if logging is dynamically disabled, each log-statement must at least check whether logging is enabled which might have a cost at runtime. The option DYNAMIC_LOG can then disable logging already at compile-time

include/phasar/PhasarLLVM/DataFlowSolver/Mono/CallString.h
```diff
@@ -67,7 +68,8 @@ template <typename T, unsigned K> class CallString {
   friend bool operator<(const CallString &Lhs, const CallString &Rhs) {
     return Lhs.cs < Rhs.cs;
   }
-  friend std::ostream &operator<<(std::ostream &OS, const CallString &C) {
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &OS,
+                                       const CallString &C) {
     std::copy(C.CS.begin(), --C.CS.end(), std::ostream_iterator<T>(OS, " * "));
```
fabianbs96, line 73: Does std::ostream_iterator really work with llvm::raw_ostream? Or does this only compile, because it is never instantiated?

include/phasar/PhasarLLVM/DataFlowSolver/Mono/Solver/InterMonoSolver.h
```diff
@@ -177,17 +176,17 @@ template <typename AnalysisDomainTy, unsigned K> class InterMonoSolver {
   }
 
   void processNormal(std::pair<n_t, n_t> Edge) {
-    std::cout << "Handle normal flow\n";
+    llvm::outs() << "Handle normal flow\n";
```
fabianbs96, line 179: Maybe, we should consider using the logger here

include/phasar/Utils/DebugOutput.h
```diff
@@ -102,9 +98,9 @@ template <typename T> struct PrettyPrinter {
 };
 
 template <typename T>
-std::ostream &operator<<(std::ostream &OS, const PrettyPrinter<T> &P) {
-  llvm::raw_os_ostream ROS(OS);
-  ROS << P;
+llvm::raw_ostream &operator<<(llvm::raw_ostream &OS,
```
fabianbs96, line 101: This function seems to be redundant

include/phasar/Utils/NlohmannLogging.h
```diff
@@ -0,0 +1,59 @@
...
+#include "llvm/Support/FormatVariadic.h"
+
+#include "nlohmann/json.hpp"
+
+namespace llvm {
+class raw_ostream;
+}
+
+namespace nlohmann {
+namespace detail {
```
fabianbs96, line 22:
```suggestion
namespace nlohmann::detail {
```

include/phasar/Utils/NlohmannLogging.h
```diff
@@ -0,0 +1,59 @@
...
+ * available under the terms of LICENSE.txt.
+ *
+ * Contributors:
+ *     Martin Mory and others
+ *****************************************************************************/
+
+#ifndef PHASAR_UTILS_NLOHMANNLOGGING_H
+#define PHASAR_UTILS_NLOHMANNLOGGING_H
+
+#include "llvm/Support/FormatVariadic.h"
```
fabianbs96, line 13:
```suggestion
#include "llvm/Support/raw_ostream.h"
```

include/phasar/Utils/NlohmannLogging.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include "nlohmann/json.hpp"
+
+namespace llvm {
+class raw_ostream;
+}
+
+namespace nlohmann {
+namespace detail {
+template <typename CharType>
```
fabianbs96, line 23: Does this really need to be a template? As far as I know, llvm::raw_ostream only supports one CharType - `char`

lib/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEGeneralizedLCA/JoinEdgeFunction.cpp
```diff
@@ -37,9 +37,9 @@ JoinEdgeFunction::JoinEdgeFunction(
     auto Ins = Seen.insert(Top);
     if (Top != nullptr && !Ins.second &&
         !dynamic_cast<AllBottom<IDEGeneralizedLCA::l_t> *>(Top)) {
-      std::cerr << "WARNING: cyclic dependency! @" << Ctr << "#";
-      Top->print(std::cerr);
-      std::cerr << std::endl;
+      PHASAR_LOG_LEVEL(WARNING, "cyclic dependency! @" << Ctr << "#");
+      Top->print(llvm::errs());
+      llvm::errs() << '\n';
```
fabianbs96, line 42: Maybe surround with `IF_LOG_ENABLED`

lib/PhasarLLVM/TaintConfig/TaintConfig.cpp
```diff
@@ -106,13 +106,13 @@ findAllFunctionDefs(const ProjectIRDB &IRDB, llvm::StringRef Name) {
       FnDefs.push_back(F);
     }
   } else if (FnDefs.size() > 1) {
-    std::cerr << "The function name '" << Name.str()
-              << "' is ambiguous. Possible candidates are:\n";
+    llvm::errs() << "The function name '" << Name.str()
```
fabianbs96, line 109:
```suggestion
    llvm::errs() << "The function name '" << Name
```

llvm::raw_ostream supports direct output of llvm::StringRef without converting to std::string;

same below

include/phasar/Utils/Logger.h
vulder: Ah, so you want to have an option to disable this check at all.
I think, we should keep this for now but maybe at some pointer later on, find a 'better' implementation maybe?
Just pasting the macro everywhere, always looks so bloated. My idea goes in the direction of using a `class` with some `consteval` magic that we can use to define the log scopes.
But anyhow, we should discuss this at some other point :D

lib/Utils/Logger.cpp
MMory: What happens if the stream cannot be opened? then no insertion takes place and true is returned ...?

lib/Utils/Logger.cpp
MMory: do they need to be moved?
also you overlooked that disable() is broken - now I didn't :D

MMory: are you saying that the compiler would always inline these functions if they were in the header?

fabianbs96: Not always, but we give it the chance then. 
If we have the implementation in the .cpp file, only that compilation unit knows the function body, so inlining (without LTO) is limited to this one CU. With the implementation in the header we give the compiler the chance to inline these functions wherever the header is included

lib/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEGeneralizedLCA/JoinEdgeFunction.cpp
fabianbs96: I wrote the initial analysis outside of phasar (recall the PG). I don't remember, who ported it to phasar

lib/PhasarLLVM/TaintConfig/TaintConfig.cpp
MMory: done, also elsewhere.

include/phasar/Utils/NlohmannLogging.h
MMory: removed.

include/phasar/PhasarLLVM/DataFlowSolver/Mono/CallString.h
MMory: don't know. @pdschubert ?

lib/Utils/Logger.cpp
MMory: it actually needed an implementation, I have deprecated this function and we will remove it in the overnext release to allow our users to migrate.

lib/Utils/Logger.cpp
MMory: done

include/phasar/Utils/Logger.h
```diff
@@ -45,45 +43,131 @@ enum SeverityLevel {
...
-#define LOG_IF_ENABLE_BOOL(condition, computation) ((void)0)
-#define LOG_IF_ENABLE(computation) ((void)0)
+#define IF_LOG_ENABLED_BOOL(condition, computation) ((void)0)
+#define PHASAR_LOG(computation) ((void)0)
+#define PHASAR_LOG_CAT(cat, message) ((void)0)
+#define PHASAR_LOG_LEVEL_CAT(level, cat, message) ((void)0)
+#define PHASAR_LOG_LEVEL(level, message) ((void)0)
 #define IS_LOG_ENABLED false
 // Have a mechanism to prevent logger usage if the code is not compiled using
 // the DYNAMIC_LOG option:
```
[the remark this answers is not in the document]
MMory: removed.

lib/DB/Hexastore.cpp
```diff
@@ -89,21 +86,21 @@ vector<HSResult> Hexastore::get(array<string, 3> EdgeQuery,
       }
     }
   }
-  string CompiledQuery =
-      str(format(QueryString) % EdgeQuery[0] % EdgeQuery[1] % EdgeQuery[2]);
+  std::string CompiledQuery = str(boost::format(QueryString) % EdgeQuery[0] %
+                                  EdgeQuery[1] % EdgeQuery[2]);
```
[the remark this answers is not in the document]
MMory: that should be in a different PR I believe.

include/phasar/PhasarLLVM/DataFlowSolver/Mono/Solver/InterMonoSolver.h
MMory: @pdschubert you agree? what log levels? should we leave that for a different PR and contribution?

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
```diff
@@ -366,12 +365,13 @@ class IDEInstInteractionAnalysisT
             } else {
               Facts.insert(Src);
             }
-            LOG_IF_ENABLE([&]() {
+            IF_LOG_ENABLED([&]() {
```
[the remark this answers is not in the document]
vulder: As long as the code is correctly scoped, this should work. It's important that the check here only get's executed once, in the no dynamic debug case.

include/phasar/PhasarLLVM/DataFlowSolver/Mono/CallString.h
vulder: Try it out?
I would think, it's just not instantiated and would not compile otherwise.

fabianbs96: I have tried it out: ostream_iterator does not work with raw_ostream. 

However, llvm::interleave works very well:
`llvm::interleave(C.CS, OS, " * ");`

vulder: This solution looks much more readable :+1:

pdschubert: Yep, looks good ;-)

pdschubert:
Thanks. I've already successfully compiled and ran it, and did some minor adjustments. Will give it a final look tomorrow morning and then merge it :+1:

PR #511, opened 2022-07-29
Redesign ProjectIRDB

fabianbs96:
The redesign of the ProjectIRDB, as discussed

PR #515, opened 2022-08-08
Redesign ICFG

fabianbs96:
Improvements to the CFG and ICFG interfaces similar to the ProjectIRDB in #511.

#511 should be merged first


pdschubert (approved):
Thanks, Fabian.

Most of this looks pretty good. But we should definitely add some performance benchmark for the various types of analysis implemented in PhASAR sometime soon. Some of the changes introduce lots of complexity and therefore, we should be able to tell whether certain changes are worth while using regression tests.

We should discuss this in our meeting on Friday.

PR #519, opened 2022-09-16
Feature/statistic export

janniclas:
I added the possibility to emit statistics gathered in the GeneralStatisticsAnalysis as a JSON object. Therefore, I added a new emitter flag and the possibility to retrieve the json object from the analysis class. The printJson method was added to the Project IRDB.


janniclas:
I didn't get the git pre commit hooks and the code formatter to run. So please ignore these failings and just review the code. I'll fix the style errors later and push an update

lib/DB/ProjectIRDB.cpp
```diff
@@ -266,6 +268,10 @@ void ProjectIRDB::print() const {
   }
 }
 
+void ProjectIRDB::printAsJson(llvm::raw_ostream &OS) const {
+  OS << StatsJson.dump();
```
fabianbs96, line 272:
```suggestion
  OS << StatsJson.dump(4) << '\n';
```

Otherwise, it may not format it properly

fabianbs96 (approved):
Looks good to me; 

Works on my machine, just the formatting should be adjusted as commented

PR #521, opened 2022-09-22
F fix cyclic dependencies

fabianbs96:
There are linking issues for phasar on some systems; however, simply adding the link libraries to the respective targets creates cyclic dependencies.
This PR aims to resolve this issue


vulder (approved):
At least for me this PR fixed the discovered linking issues.
thx

pdschubert (approved):

PR #524, opened 2022-10-17
Feature/statistics update

janniclas:
This merge request contains

-  typo fixes `statistic -> statistics`
- added the module name to the exported json statistics, to easily identify the result files of multiple analyses.
- Update to the Docker development env file to correctly set the `PATH` variable
- Updates to the Dockerfile to let it build way faster ( we don't build LLVM in the image, because this image is intended to run analyses in release mode and from my understanding we don't need the custom build version of LLVM for this). This update was inspired from the test setup from PhASAR in the GitHub actions.


janniclas:
Extended the request with a new GitHub action to build the docker image and publish it to GitHubs container registry. This can greatly ease the initial adoption of PhASAR imho.
It would be even better if we also push the image to the official docker registry, if you think this is desirable just reach out to me and I can walk you through the necessary steps @pdschubert .

fabianbs96 (approved):
The C++ part of the PR looks good to me; the docker.yml I cannot verify due to missing expertise

pdschubert (approved):

pdschubert:
Hey @vulder can you tell why the pre-commit fails here? And if so, how to fix it?

Thanks!
Philipp

vulder:
@pdschubert The hook most often tells you the issue: `trim trailing whitespace.................................................Failed`
After the docker line are trailing whitespaces, remove them and the check will be happy again :grinning:

.github/workflows/docker.yml
```diff
@@ -0,0 +1,52 @@
+name: Docker Image
+
+on:
+  push:
+    branches: [ development ]
+  pull_request:
+    branches: [ development ]
```
vulder, line 7: @pdschubert  
I'm not sure if you really want that. From what I know, this means every PR (before approval and merging) will build a new docker image an push it into the registry. Even for every new commit on a open PR this will build and upload a new image.

janniclas: oh sorry that is a leftover from testing the change to do the multi platform build. I'll remove this in a moment when fixing the whitespaces.
sorry testing CI updates is always painful and I needed a deployed image to verify if it actually works on multiple platforms

PR #525, opened 2022-10-20
Fixes linking issues in utils and tests

vulder:
* phasar_utils requires boost libs and std::filesystem
* operator<< for SpecialMemberFunctionType was not fully ported to llvm streams

### DIFF
--- a/lib/PhasarLLVM/ControlFlow/CFG.cpp
+++ b/lib/PhasarLLVM/ControlFlow/CFG.cpp
@@ -7,7 +7,9 @@
  *     Philipp Schubert and others
  *****************************************************************************/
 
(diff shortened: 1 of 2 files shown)


fabianbs96 (approved):
Looks good to me; 

Maybe, we should not implement free functions inside a namespace (e.g. psr). To make such problems visible at compile-time, we may prefer implementing such functions on global scope and explicitly qualify them with the full namespace, e.g. 
``` c++
llvm::raw_ostream &psr::operator<<(llvm::raw_ostream &OS, 
                                   const SpecialMemberFunctionType &SMFT) {
```

pdschubert (approved):

PR #527, opened 2022-10-26
Update `master`

pdschubert:


janniclas:
while at it, should we rename this branch to main before we start using it again?

PR #528, opened 2022-10-26
Fix IDE edge functions

fabianbs96:
`AllBottom::composeWith` did not correctly model function composition semantics in most of phasar's IDE analyses. This PR fixes this.
For future IDE implementations keep the following insights in mind:
- On $Λ \rightarrow Λ$ edges always attach the EdgeIdentity
- For any edge functions $f$ and $g$ and edge value $l$ it should hold: $f$-> $composeWith(g)$-> $computeTarget(l) == g$-> $computeTarget(f$-> $computeTarget(l))$, especially:
- Any edge function $f$ composed with AllBottom should result in AllBottom
- Any edge function $f$ composed with EdgeIdentity should result in $f$
- AllBottom should only be composed with AllBottom or EdgeIdentity, otherwise there is a bug in the analysis
- Policies such as F::composeWith(G) calling G::composeWith(F) are bad and lead to hard-to-find bugs


pdschubert:
Is this still a draft or can I start reviewing it?

pdschubert (approved):

PR #530, opened 2022-10-30
Reorder Analysis Dependencies

fabianbs96:
The current definition of IFDSTabulationProblem and IDETabulationProblem has two drawbacks that are addressed in this PR:
- IDETabulationProblem inherits from IFDSTabulationProblem which requires the IFDSSolver to create a IfdsToIDETabulationProblem wrapper for propagating it to the generic IDESolver. IN thie PR, we flip the inheritance relation making IDETabulationProblem a superclass of IFDSTabulationProblem avoiding the mentioned issue.
- Each IfdsIde analysis problem requires (through their superclass) the complete set of IRDB, ICFG, PT and TH, from which most analyses only need the IRDB. Especially the PT computation oftentimes leads to a significant performance overhead, so we want to skip it, if we don't need it. However, as of now, deciding which analysis problem really needs which helper analysis requires in-depth knowledge of the implementation details of the analyses in question. In this PR, we let the IDE/IFDS TabulationProblem only require the IRDB such that each client analysis can declare itself, which helper analyses it needs


pdschubert:
Thanks, @fabianbs96. I'll review this through the next few days :+1:

vulder:
One quick question before hand. Isn't `IFDS <: IDE` conceptually the wrong direction? As a IDE problem is an IFDS one but not the other way around? Maybe a template structure where the "base parts" are IFDS and IDE parts get's woven in with a bit template magic. Could need a small adaption to the IDESolver, but maybe lead to a cleaner interface?

PS: I'm not sure, I'm just asking questions to explore design possibilities.

I'll also take a peak at the code later ;)

fabianbs96:
Thanks @vulder for this advice; I was also thinking about that. However, as the IDESolver is designed as a base-class of the IFDSSolver it only makes sense for the IDEProblem to be a base-class of the IFDSProblem. 
I aggree with you that conceptually it would make more sense for the IFDS _Solver_ to be the base of the IDESolver. The reason I changed it this way now is that I didn't want to do larger changes on the IDESolver.
Anyway I would really like to see a code peak from you.

vulder:
:+1: For the code, can you give me a few pointers to the important files, so I don't have to look at all the dependent changes :laughing:

fabianbs96:
The most important files are the IDE- and IFDSTabulationProblem.h and maybe the second ctor of the IDESolver. I also did quite a lot changes in the AnalysisController, but these changes should not be that important ona conceptual level.
To see how the changes affect analysis problems you may look at IFDSTaintAnalysis.h/.cpp

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
```diff
@@ -205,16 +205,14 @@ class IDEInstInteractionAnalysisT
       std::variant<n_t, const llvm::GlobalVariable *> InstOrGlobal);
 
   IDEInstInteractionAnalysisT(
-      const ProjectIRDB *IRDB, const LLVMTypeHierarchy *TH,
-      const LLVMBasedICFG *ICF, LLVMPointsToInfo *PT,
+      const ProjectIRDB *IRDB, const LLVMBasedICFG *ICF, LLVMPointsToInfo *PT,
```
vulder, line 208: Do I understand this correctly, `TH` now is no longer needed on the Problem by the Solver takes care of this?

And for the `ICFG` and `PT` we store the locally?

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Solver/IDESolver.h
```diff
@@ -54,37 +36,29 @@
...
 
-namespace psr {
-
-// Forward declare the Transformation
-template <typename AnalysisDomainTy, typename Container>
-class IFDSToIDETabulationProblem;
+#include "nlohmann/json.hpp"
 
-struct NoIFDSExtension {};
+#include "boost/algorithm/string/trim.hpp"
```
vulder, line 41: Maybe on the fly get rid of `boost::trim` and replace it with `llvm::StringRef::trim`?

vulder (commented):
With regard to the second IDESolver ctor, is it not just gone?

include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/IDEInstInteractionAnalysis.h
fabianbs96: Yes, in fact TH is not used at all by the analysis problems, so there is no need to them to store a reference to it. As of now, the only use of TH is in the call-graph resolver inside LLVMBasedICFG.. The PT and ICFG are used in some analysis problems, but by far not all, so only those problems needing them will recieve a pointer. Previously, all analysis problems needed to store all of TH, PT and ICFG -- and then not using them. In future, we might want to only compute that subset of {TH, PT, ICFG} that is really needed.

The solver only needs the problem and the ICFG, nothing more. Previously, the solver retrieved its ICFG from the problem; now it gets it directly as the problem may not have access to an ICFG. Inside the solver, there is no difference at all, as it directly stores a pointer to the ICFG anyway.

vulder: LGTM

MMory (approved):
lgtm

PR #546, opened 2022-12-03
Extended calculated statistics

janniclas:
This PR adds statistics for the following instruction counts:

-  size_t Branches = 0;
- size_t GetElementPtrs = 0;
- size_t Terminators = 0;
- size_t PhiNodes = 0;

Further some dead code was removed.


include/phasar/PhasarLLVM/Passes/GeneralStatisticsAnalysis.h
```diff
@@ -45,6 +45,10 @@ class GeneralStatistics {
   size_t LoadInstructions = 0;
   size_t MemIntrinsics = 0;
   size_t GlobalPointers = 0;
+  size_t Branches = 0;
+  size_t GetElementPtrs = 0;
+  size_t Terminators = 0;
```
fabianbs96, line 50: It seems to be redundant to count terminators as there is always exactly one terminator per basic-block

fabianbs96 (commented):
Looks already quite good; 
maybe you want to print the extended statistics from phasar-llvm -S as well

include/phasar/PhasarLLVM/Passes/GeneralStatisticsAnalysis.h
janniclas: good point, didn't think about that.

janniclas (commented):
removed separate counter for terminators

janniclas:
@fabianbs96 I included the changes we discussed in regards of exposing the complete GeneralStatistics from the IRDB class, as well as overloading the << operator in the statistics class to ease printing and cleanup the code in phasar-llvm.cpp

PR #550, opened 2022-12-10
Cleanup FlowFunction Templates

fabianbs96:
PhASAR provides a number of predefined templates implementing the FlowFunction interface. These templates are especially useful to quickly define data flows in a clean way. However, these flow function templates are hard to use, hard to find and not documented at all which leads analysis developers not to use them.

This PR comes with the following two contributions:
 1. The flow function templates are rewritten such that they are easier to use and require less boilerplate in the actual analysis definition
 2. The flow function templates are documented with code comments

As a proof of concept, the IDELinearConstantAnalysis has been rewritten to use the new flow function templates.


include/phasar/PhasarLLVM/DataFlowSolver/IfdsIde/LLVMFlowFunctions.h
```diff
@@ -10,413 +10,424 @@
...
+/// all dataflow facts x holding at exit:
+///   f(rv) = {r}   if PropagateParameter(rv, rv) else {},
+///   f(px) = {ax}  if PropagateParameter(px, px) else {},
+///   f(g)  = {g}   if PropagateGlobals else {},
+///   f(Λ)  = {Λ}   if PropagateZeroToCaller else {},
+///   f(x)  = {ax}  if PropagateParameter(ax, x) else {}.
+///
+/// NOTE: PropagateParameter is used to dermine if both output parameters and
+/// the return value should be propagated back to the caller. So, it may make
+/// sense to provide two different predicates combined by psr::Overloaded.
```
janniclas, line 238: I think it might make the analysis code even more readable and this function easier to use if we explicitly expose two functions one for each case, instead of this general form.

fabianbs96: done

janniclas:
I'm happy with the changes from a usability standpoint. However, a final judgement of the improvements can only be done after working with the new structure, but I'm very happy to try it out so I suggest to merge it :-)

@MMory could you review the changes from a technical view point?

MMory (approved):
lgtm, we can decide about removing deprecated FFs at a later point.

PR #551, opened 2022-12-11
Rework Pointer- and Alias Analysis Interfaces

fabianbs96:
PhASAR provides its own abstractions for analyses of pointer relations via the PointsToInfo interface. However, none of the implementations of this interface compute points-to information. They all provide *alias* information. Hence, the name PointsToInfo is misleading (should better be AliasInfo).

This PR has two contributions:
1. Rename `PointsToInfo` to `AliasInfo` and add a new interface `PointsToInfo` for future points-to analysis implementations.
2. Bring the AliasInfo and PointsToInfo interfaces to the new design that is already applied for CFG, ICFG and ProjectIRDB (in its own branch)

Note: #511 should be merged first.


MMory (approved):
lgtm, will merge once CI works again.

MMory:
This is a breaking change, isn't it? Before merging BreakingChanges.md should be updated then.

fabianbs96:
Yes, this one is breaking

MMory (approved):

PR #570, opened 2023-01-17
Fix LLVM Issue in CI

fabianbs96:


fabianbs96:
Thanks @vulder for pointing out this fix: https://github.com/se-sic/vara-feature/commit/04bbe3f4968e9b3dbc9b40b9350757cab617afc0

MMory (approved):
We need to revert this once LLVM (hopefully) fixes their screw-up.

PR #571, opened 2023-01-18
Fix Dockerfile

fabianbs96:
Apply the CI fix to the Dockerfile


MMory (approved):
We need to revert this once LLVM finally fix their screw-up.

PR #572, opened 2023-01-18
Simplify Analysis Construction

fabianbs96:
With the Reorder Analysis Dependencies PR the implementations of the IFDS and IDETabulationProblem now have different constructor requirements regarding which helper analyses they need. With the introduction of the `HelperAnalyses` type, analysis construction can be simplified again.
However, we don't want to force users to add another construction to their analysis problem description taking a `HelperAnalyses&` as parameter.

This PR thus adds a factory function `createAnalysisProblem` that statically determines, which helper analyses are needed and automatically invokes the right ctor.
This factory function has been integrated into all unittests and the AnalysisController


include/phasar/PhasarLLVM/Utils/Printer.h
```diff
@@ -140,12 +101,6 @@ template <typename AnalysisDomainTy> struct FunctionPrinter {
 };
 
 template <typename ContainerTy> struct ContainerPrinter {
-  ContainerPrinter() = default;
-  ContainerPrinter(const ContainerPrinter &) = delete;
-  ContainerPrinter &operator=(const ContainerPrinter &) = delete;
-  ContainerPrinter(ContainerPrinter &&) = delete;
-  ContainerPrinter &operator=(ContainerPrinter &&) = delete;
-  virtual ~ContainerPrinter() = default;
```
MMory, line 148: should remain

MMory (approved):
lgtm, thanks!

PR #573, opened 2023-01-18
Simplify InstInteractionAnalysis

fabianbs96:
The IDEInstInteractionAnalysis is one of the most complex analyses within PhASAR. So, it is a good candidate for applying the new flow function templates as well as general cleanup.

@vulder: Does the analysis still work in your environment?


vulder (approved):
LGTM
all test are still passing on our side

PR #574, opened 2023-01-18
Rework Folder Structure

fabianbs96:
While PhASAR has grown over the years, the initial folder structore does no longer correspond to the semantics.
For example, non-LLVM stuff is placed within PhasarLLVM, etc.

This PR moves files between folders in order to re-sync file- and folder- semantics


include/phasar/DataFlow/IfdsIde/InitialSeeds.h
```diff
@@ -72,39 +72,6 @@ template <typename N, typename D, typename L> class InitialSeeds {
...
-      for (const auto &[Fact, Value] : Facts) {
-        OS << "> ";
-        printFact(Fact);
-        OS << " --> \\." << Value << "\n";
-      }
-      OS << "\n";
-    }
-    OS << "========================== End Seeds ==========================\n";
-  }
-
```
MMory, line 107: not sure whether I would remove this even if dead code, might come handy for debugging purposes?

include/phasar/DataFlow/Mono/IntraMonoProblem.h
```diff
@@ -66,18 +65,14 @@ class IntraMonoProblem : public NodePrinter<AnalysisDomainTy>,
...
-        EntryPoints(std::move(EntryPoints)) {
-    static_assert(is_cfg_v<c_t, AnalysisDomainTy>,
-                  "c_t must implement the CFG interface!");
-    static_assert(std::is_base_of_v<ProjectIRDBBase<db_t>, db_t>,
-                  "db_t must implement the ProjectIRDBBase interface!");
-  }
+        EntryPoints(std::move(EntryPoints)) {}
 
-  ~IntraMonoProblem() override = default;
+  virtual ~IntraMonoProblem() = default;
```
MMory, line 75: needs to remain override.

lib/PhasarLLVM/Utils/phasar_llvm_utils-config.cmake
```diff
@@ -3,7 +3,10 @@ set(PHASAR_phasarllvm_utils_COMPONENT_FOUND 1)
 list(APPEND
   PHASAR_LLVM_DEPS
   Support
-  Core)
+  Core
+  BitWriter
+  Demangle
+)
```
MMory, line 9: for some reason here we still have this file, in other places we don't. forgot this one?

lib/Utils/Utilities.cpp
```diff
@@ -77,20 +114,22 @@ const llvm::Type *stripPointer(const llvm::Type *Pointer) {
...
+  if (Name.startswith("_D")) {
+    // D ABI
+    return true;
   }
-  return OS;
+  // Microsoft ABI is a bit more complicated...
+  return Name != llvm::demangle(Name.str());
 }
 
 bool StringIDLess::operator()(const std::string &Lhs,
```
MMory, line 135: please make a seperate PR for this change, this is unrelated to folder structure and too complex just to sneak it in

MMory:
might interfere a bit with #516 
@vulder would be nice if you could test whether the in-tree build of this change is fine.

lib/Utils/Utilities.cpp
janniclas: Agreed, and tag me in the PR please I can add the stuff regarding the Swift ABI and mangling :)

Here we want to essentially check for "begins with '$s'".

vulder (approved):
LGTM

Currently, with a few tweaks I could get this PR to work.

What I needed to change is shown here [IntegrationPR](https://github.com/se-sic/VaRA/pull/988/files) maybe some of these should be added to the dev suggestions part.

PR #575, opened 2023-01-18
Cleanup Configuration

fabianbs96:
Detect relevant phasar directories at compile-time instead of runtime. This has two major advantages:
- No need to keep PhasarConfig and the CMakeLists.txt in sync regarding (config) directories
- No weird initialization orders of global variables


CMakeLists.txt
```diff
@@ -120,14 +120,18 @@ else()
   set(PHASAR_STD_FILESYSTEM stdc++fs)
 endif()
 
-option(PHASAR_CONFIG_INSTALL_PREFIX "Customizes the directory, where configuration files for PhASAR are installed (default is $(HOME)/.config)")
+set(PHASAR_CONFIG_INSTALL_PREFIX "" CACHE STRING "Customizes the directory, where configuration files for PhASAR are installed (default is $(HOME)/.config)")
 
-if(${PHASAR_CONFIG_INSTALL_PREFIX} STREQUAL "")
+if("${PHASAR_CONFIG_INSTALL_PREFIX}" STREQUAL "")
   set(PHASAR_CONFIG_INSTALL_DIR "$ENV{HOME}/.config/phasar")
 else()
```
MMory, line 127: (documenting our discussion about this PR on 9 Feb) this hardcodes the building user's home directory in the binary. The config load path should be determined at runtime. there should be a system-wide location for configuration, such as /opt/phasar/config, to which configurations are installed. Users should have the possibility to provide their config file in ~/.config/phasar if wished. I would go for KISS and abort the search for the config file when it is found in the home directory path. We should abstract away the search for a particular config file with a function like getPhasarConfigFilename(std::string &Filename)

lib/Config/Configuration.cpp
```diff
@@ -36,47 +43,108 @@ PhasarConfig::PhasarConfig() {
   SpecialFuncNames.insert({"_Znwm", "_Znam", "_ZdlPv", "_ZdaPv"});
 }
 
+std::optional<llvm::StringRef>
+PhasarConfig::LocalConfigurationDirectory() noexcept {
+  static std::string DirName = []() -> std::string {
+    const auto *Home = getenv("HOME");
```
MMory, line 49: Fabian, you mentioned that llvm might have a function for this?

lib/Config/Configuration.cpp
```diff
@@ -36,47 +43,108 @@ PhasarConfig::PhasarConfig() {
...
     SpecialFuncNames.insert({"_exit"});
+    return;
   }
+
+  llvm::SmallVector<llvm::StringRef, 0> GlibcFunctions;
+  llvm::SplitString(*Glibc, GlibcFunctions, "\n");
+
+  llvm::transform(GlibcFunctions,
+                  std::inserter(SpecialFuncNames, SpecialFuncNames.end()),
+                  [](llvm::StringRef Str) { return Str.trim().str(); });
```
MMory, line 120: this is pretty similar to the parsing below. although just some lines, I think it might make sense to create a function for this.

MMory (commented):
lgtm, except minor

MMory (approved):
lgtm, thank you!

PR #584, opened 2023-02-13
fix: make the OpenSSL-based unittests compile

jusito:
OpenSSL tests were disabled during latest pr.


MMory (approved):
lgtm

MMory:
just wonder why they were disabled, don't recall that. @fabianbs96?

PR #585, opened 2023-02-14
intrinsicBinaryOp, GetElementPtr, and ExtractValue support for IDE LCA

janniclas:
This PR adds flow and edge functions to the IDE_LCA implementation to handle getelementptr and extractvalue instructions. 
These oftentimes appear in Swift-based LLVM IR and with the added changes we are able to handle the following constructs:

```
%8 = call { i64, i1 } @llvm.sadd.with.overflow.i64(i64 %0, i64 41), !dbg !79
 %9 = extractvalue { i64, i1 } %8, 0, !dbg !79
 %10 = extractvalue { i64, i1 } %8, 1, !dbg !79
```

```
%TSi = type <{ i64 }>
%4 = alloca %TSi, align 8
%8 = call { i64, i1 } @llvm.sadd.with.overflow.i64(i64 %0, i64 41), !dbg !79
%._value1 = getelementptr inbounds %TSi, %TSi* %4, i32 0, i32 0, !dbg !79
store i64 %9, i64* %._value1, align 8, !dbg !79
```

Additionally, I sneaked in a bugfix for printing the results of the LCA.


fabianbs96:
@janniclas I have rewritten the "look-through GEP" for the StoreInst; can you check whether your test cases still work?

janniclas:
Thx for the cleanup !
I'll checkout how to add and run test cases with PhASAR tomorrow morning. Then I'll just include my tests here. Meant to do this anyway. However, I'll need to provide the generated LLVM IR so you don't have to have setup the swift compiler to compile the test cases..

janniclas:
@fabianbs96 Nice find with stripPointerCasts ! works as expected :)

I added the test cases and the corresponding IR. However, as stated in the commit message, we currently don't compile the Swift test case to IR, so we have to manually copy the IR file I included to the fitting test directory. 
I bet cmake can do this for me but I have no idea how - @jusito  maybe you could help? What I want to do for now is copy every *_swift.ll file in the test directories to build/test/...

janniclas:
This should be merged after https://github.com/secure-software-engineering/phasar/pull/587

unittests/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/CMakeLists.txt
```diff
@@ -24,7 +24,20 @@ else()
...
+	IFDSConstAnalysisTest.cpp
+	IFDSTaintAnalysisTest.cpp
+	IDEInstInteractionAnalysisTest.cpp
+	IDELinearConstantAnalysisTest.cpp
+	IDELinearConstantAnalysis_DotTest.cpp
+	IFDSUninitializedVariablesTest.cpp
+	IDEGeneralizedLCATest.cpp
+	IDEExtendedTaintAnalysisTest.cpp
+	IDETSAnalysisFileIOTest.cpp
+	)
```
MMory, line 40: Fabian suggests another variable like IfdsIdeSwiftProblemSources that would be filled with the swift tests optionally. This if/else is a bit error prone and redundant.

MMory (commented):
lgtm, except minor

unittests/PhasarLLVM/DataFlowSolver/IfdsIde/Problems/CMakeLists.txt
janniclas: Makes sense. If you are fine with it we could already merge it. I'll add many Swift test cases later this week and during that process I have to edit this file anyway and will include this feedback then.

MMory (approved):

PR #586, opened 2023-02-14
Fix Callgraph

fabianbs96:
The CallGraph contains callees where the function signatures do not match the call-site. 
This causes some assertion failures in flow functions.


MMory (approved):
lgtm, thanks!

PR #587, opened 2023-02-15
Including Test Setup for Swift

janniclas:
With this PR I added the possibility to compile Swift test cases to LLVM IR.
The idea is to include this as an optional setting with default to set OFF, so our users can decide whether they want to run those locally. However, I would like to include the tests in our build pipeline to make sure that we are notified if changes affect the Swift-based LLVM IR.


janniclas:
@jusito feedback appreciated :)

janniclas:
The pr is ready to be merged. 
It is expected that the unit test fails since the extension to the LCA fixing the test case will be merged after this one.

test/llvm_swift_test_code/linear_constant/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+#  cmake_minimum_required(VERSION 3.21)
+# project(P
+#   LANGUAGES C CXX Swift)
```
jusito, line 3: please remove unused code

unittests/TestUtils/TestConfig.h
```diff
@@ -18,6 +18,9 @@ inline const std::string PathToTxtTestFiles(PhasarConfig::PhasarDirectory() +
 inline const std::string PathToJSONTestFiles(PhasarConfig::PhasarDirectory() +
                                              "test/json_test_code/");
 
+inline const std::string PathToSwiftTestFiles(
+    PhasarConfig::PhasarDirectory() + "build/test/llvm_swift_test_code/");
```
jusito, line 22: I would prefer not having build directly here, because it assumes the build folder is always called like this. Which isn't the case e.g. for cmake presets, but in TestConfig.h we are assuming it already, so...

jusito (changes requested):

unittests/TestUtils/TestConfig.h
janniclas: Tbh, I just copied the existing setup for the other tests. So would it be fine for you to keep it this way? We can rework the whole setup for all tests with a different PR, this would be a bit off topic for this one I guess?

jusito: sure

MMory:
This looks pretty overlapping to #585 , do we still need/want this?

janniclas:
> This looks pretty overlapping to #585 , do we still need/want this?

I intentionally splitted it in two PRs so it's easier to review the changes. This one just includes the changes to the test setup, the other one changes to the analysis + new tests utilizing this new test setup.

MMory:
closing this according to private conversation with @janniclas

PR #590, opened 2023-02-22
Draft: Fix RPATH for install

fabianbs96:
The install does not install libraries into `${CMAKE_INSTALL_PREFIX}/lib/phasar`, but directly into `${CMAKE_INSTALL_PREFIX}/lib`.
So, the linker rpath should point there as well


vulder:
@MMory What is the desired location when you package phasar? The RPATH should be correct, however, I'm not sure if you want to directly install into `lib`, so I'll just wanted to point that out.

MMory:
For testing, I have set `CMAKE_INSTALL_PREFIX` to `/opt/phasar`. When doing so, all of phasars libs are installed to /opt/phasar/lib64, but the schema validator is installed to /opt/phasar/lib. The RPATH is set to /opt/phasar/lib64, leaving the schema validator unresolved. lib seems to be hardcoded on the schema validator side. @fabianbs96 what do you think we should do?

fabianbs96:
Thanks @MMory for that info. That is indeed an issue. 
I marked this PR as draft again, as it seems there is more work to be done. 
We should also think about whether it is a good idea to install external dependencies at all

CMakeLists.txt
```diff
@@ -127,7 +131,7 @@ else()
   set(PHASAR_STD_FILESYSTEM stdc++fs)
 endif()
 
-set(PHASAR_CONFIG_INSTALL_DIR "/opt/phasar/config/" CACHE STRING "Customizes the directory, where configuration files for PhASAR are installed (default is /opt/phasar/config)")
+set(PHASAR_CONFIG_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/.phasar-config/" CACHE STRING "Customizes the directory, where configuration files for PhASAR are installed (default is /usr/local/.phasar-config)")
```
vulder, line 134:
```suggestion
set(PHASAR_CONFIG_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/.phasar-config/" CACHE STRING "Customizes the directory, where configuration files for PhASAR are installed (default is /usr/local/.phasar-config)" FORCE)
```
If we don't for the setting of the variable the cached state will stay the same, hence a change to `CMAKE_INSTALL_PREFIX` will not override the variable.

vulder (approved):

PR #593, opened 2023-02-24
Minor update to IR Statistics

janniclas:
Including a counter for global constants (seems like u were right @fabianbs96 , most of the globals generated in Swift-based LLVM IR are consts and can be ignored for most cases :) )

### DIFF
--- a/lib/PhasarLLVM/Passes/GeneralStatisticsAnalysis.cpp
+++ b/lib/PhasarLLVM/Passes/GeneralStatisticsAnalysis.cpp
@@ -130,6 +130,9 @@ GeneralStatistics GeneralStatisticsAnalysis::runOnModule(llvm::Module &M) {
       ++Stats.GlobalPointers;
     }
     ++Stats.Globals;
+    if (Global.isConstant()) {
+      ++Stats.GlobalConsts;
+    }
   }
   // register stuff in PAMM
   // For performance reasons (and out of sheer convenience) we simply initialize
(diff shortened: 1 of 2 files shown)


lib/PhasarLLVM/Passes/GeneralStatisticsAnalysis.cpp
```diff
@@ -244,6 +252,8 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &OS,
             << "LLVM IR instructions:\t" << Statistics.Instructions << "\n"
             << "Functions:\t" << Statistics.Functions << "\n"
             << "Global Variables:\t" << Statistics.Globals << "\n"
+            << "Global Variable Consts:\t" << Statistics.GlobalConsts << "\n"
+            << "Global Pointers:\t" << Statistics.GlobalPointers << "\n"
```
fabianbs96, line 256: This seems to be redundant. All globals are of pointer type by construction

fabianbs96 (approved):
lgtm, except minor

lib/PhasarLLVM/Passes/GeneralStatisticsAnalysis.cpp
janniclas: agreed. however, the code to set these counters where here before and I didn't want to remove it because I was not sure if someone uses it with exactly this output. 
```
    if (Global.getType()->isPointerTy()) {
      ++Stats.GlobalPointers;
    }
    ++Stats.Globals;
```

We can definitely remove this but I didn't want to sneak it in here to keep it clear, I just added it to the output to make it more transparent that this exists.

PR #594, opened 2023-03-01
Fix Configuration Header

fabianbs96:
The `Configuration.h` header requires the compilation symbols `PHASAR_DIR` and `PHASAR_CONFIG_DIR` to be defined. This is fine for compiling phasar itself; however when using this header from clients that just _use_ phasar as a library, these symbols may not be defined. We should not require them to be defined.

This PR fixes this by moving the parts that depend on these symbols to the `Configuration.cpp` source file.


MMory (approved):
lgtm, thanks!

PR #596, opened 2023-03-03
Fix calling llvm::Bitvector::getData() on empty BV

MMory:


MMory (commented):
lgtm

vulder (changes requested):
I'd add a test to `unittests/Utils/BitVectorSetTest.cpp` to prevent this from reoccurring :wink: 

Otherwise, this LGTM
thanks for fixing

PR #597, opened 2023-03-13
Fix Crash in GlobalCtorDtor Handling

fabianbs96:
To handle global constructors and destructors we are generating wrapper functions that call these. In some (rare) cases the generated call-sites was incompatible with the expected callee-signature (some missing pointe-bitcasts).
As this is only checked with debug-assertions on LLVM-side, this error only shows up when using a debug-build of LLVM (for example in the in-tree build).

This PR fixes the issue on phasar side.

### DIFF
--- a/lib/PhasarLLVM/ControlFlow/LLVMBasedICFGGlobalsImpl.cpp
+++ b/lib/PhasarLLVM/ControlFlow/LLVMBasedICFGGlobalsImpl.cpp
@@ -136,7 +136,14 @@ static llvm::Function *createDtorCallerForModule(
 
   for (auto It = RegisteredDtors.rbegin(), End = RegisteredDtors.rend();
        It != End; ++It) {
-    IRB.CreateCall(It->first, {It->second});
+    auto FunCallee = It->first;
+    assert(FunCallee.getFunctionType()->getNumParams() == 1);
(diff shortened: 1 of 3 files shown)


lib/PhasarLLVM/ControlFlow/LLVMBasedICFGGlobalsImpl.cpp
```diff
@@ -136,7 +136,14 @@ static llvm::Function *createDtorCallerForModule(
...
   for (auto It = RegisteredDtors.rbegin(), End = RegisteredDtors.rend();
        It != End; ++It) {
-    IRB.CreateCall(It->first, {It->second});
+    auto FunCallee = It->first;
+    assert(FunCallee.getFunctionType()->getNumParams() == 1);
+    auto *ExpectedArgType = FunCallee.getFunctionType()->getParamType(0);
+    auto *Arg = It->second;
+    if (Arg->getType() != ExpectedArgType) {
+      Arg = IRB.CreateBitOrPointerCast(Arg, ExpectedArgType);
+    }
```
vulder, line 145: Can we be sure that the call is correctly matched? Otherwise, we could introduce wrong cast here.

vulder (approved):
Except for my comment, this looks fine from the analysis side.

PR #598, opened 2023-03-18
Disables warnings in nlohmann_json

vulder:
Warnings, such as, warning: default label in switch which covers all enumeration values, from nlohmann_json spam during compilation, especially for LLVM builds. A default label for a switch is unreachable should all cases covered. LLVM prefers to warn in cases where not all cases are covered. As we cannot change the code in the json lib, we should at least disable the flood of warnings.


fabianbs96 (approved):
lgtm

PR #600, opened 2023-03-18
Fix and Simplify Initial Seeds

fabianbs96:
The `initialSeeds()` for most IFDS/IDE problems were set on the first instruction of the `EntryPoints` functions. However, the `LLVMBasedCFG` (correctly) skips debug intrinsics by default which may prevent the IDESolver from computing edge values on entry-functions where the first instruction (that one where the seeds are set) is a debug instruction.

This PR fixes this issue and additionally provides some utilities that make it easier creating correct seeds.

### DIFF
--- a/include/phasar/PhasarLLVM/ControlFlow/LLVMBasedCFG.h
+++ b/include/phasar/PhasarLLVM/ControlFlow/LLVMBasedCFG.h
@@ -49,10 +49,11 @@ public:
   }
 
 protected:
-  LLVMBasedCFGImpl(bool IgnoreDbgInstructions = true) noexcept
+  LLVMBasedCFGImpl() noexcept = default;
+  LLVMBasedCFGImpl(bool IgnoreDbgInstructions) noexcept
       : IgnoreDbgInstructions(IgnoreDbgInstructions) {}
 
-  bool IgnoreDbgInstructions = false;
+  bool IgnoreDbgInstructions = true;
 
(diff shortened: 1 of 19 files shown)


include/phasar/PhasarLLVM/ControlFlow/LLVMBasedCFG.h
```diff
@@ -49,10 +49,11 @@ template <typename Derived> class LLVMBasedCFGImpl : public CFGBase<Derived> {
   }
 
 protected:
-  LLVMBasedCFGImpl(bool IgnoreDbgInstructions = true) noexcept
+  LLVMBasedCFGImpl() noexcept = default;
+  LLVMBasedCFGImpl(bool IgnoreDbgInstructions) noexcept
       : IgnoreDbgInstructions(IgnoreDbgInstructions) {}
 
-  bool IgnoreDbgInstructions = false;
+  bool IgnoreDbgInstructions = true;
```
MMory, line 56: are there more locations that configure this behaviour? Is it the right thing to make this change in this PR?

fabianbs96: That was already the default, because the ctor parameter `IgnoreDbgInstructions` was set to true by default. THis change just makes it clearer what the actual default really is

MMory (approved):

PR #608, opened 2023-04-03
Consume solver results

fabianbs96:
The IDESolver has been found to consume a lot of memory, most of it being intermediate results. 

This PR aims at reducing the total memory overhead with two contributions:
1. The default solver configuration no longer enables recording the ESG edges in an additional map (this feature is anyway likely to be deprecated and replaced in the future). This has to be enabled explicitly. Note: `phasar-cli` already disables recordEdges by default
2. The solver results can be moved out of the IDESolver. This allows destructing the solver while retaining the solver results for subsequent analysis. To simplify the process new functions `solveIDEProblem` and `solveIFDSProblem` have been added


MMory (approved):
lgtm, thanks!

PR #609, opened 2023-04-13
Fix Logger Macros

fabianbs96:
The logger macros `PHASAR_LOG*` currently only work if used within the namesapce `psr` or in an environment that uses `using namespace psr`. 
This PR fixes that issue.


MMory (approved):
lgtm, thanks!
However, seeing this I wonder whether at some point there could be a preprocessor-free solution. Then, we would not necessarily litter the top-level namespace with some logger identifiers and could do something like psr::IF_LOG_ENABLED_BOOL or so.

PR #610, opened 2023-04-14
Update LLVMBasedAliasAnalysis

fabianbs96:
LLVM has removed their CFL based alias analysis implementations (CFLAndersAA and CFLSteensAA -- https://reviews.llvm.org/D139703), so we should remove their uses as well. This also removes some bugs due to known crashes of both CFL AA for some larger target programs.


vulder:
@fabianbs96 @MMory you should discuss internally what's a good way forward from this. Loosing these AA algorithms reduces phasars precision, so we probably need some form of replacement algorithms that ship with phasar.

fabianbs96:
As agreed we won't merge this PR until we have a proper replacement for the CFLAndersAA

PR #612, opened 2023-04-19
Fix Boost Installation

fabianbs96:
The `BOOST_VERSION` check inside `bootstrap.sh` returns exit code 1 if boost is not already installed preventing the script from continuing.


MMory (approved):
lgtm, thanks

PR #614, opened 2023-05-03
Ref-counted ModuleSlotTracker

fabianbs96:
Phasar currently crashes when the same `llvm::Module` is managed by multiple `LLVMProjectIRDB` instances at the same time. This situation occurrs for example when phasar is used in LLVM's pass infrastructure and multiple phasar-passes are run in one `opt`-call.

Reference-counting the cached ModuleSlotTracker instances fixes this problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ bin/*
 # build directories for cmake
 build/
 build_*/
+build-*/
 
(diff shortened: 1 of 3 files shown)


MMory (approved):
lgtm, thanks!

PR #617, opened 2023-05-07
Deserialization for LLVMBasedICFG

fabianbs96:
Add support for deserializing an LLVMBasedICFG from json. For that to work properly the LLVMBasedICFG::getAsJson() function has been refactored -- the result format is different now.

Additionally fixed psr::fromMetaDataId that is required for both LLVMBasedICFG- and LLVMAliasSet deserialization.

### DIFF
--- a/lib/PhasarLLVM/Pointer/LLVMAliasSet.cpp
+++ b/lib/PhasarLLVM/Pointer/LLVMAliasSet.cpp
@@ -100,7 +100,7 @@ LLVMAliasSet::LLVMAliasSet(LLVMProjectIRDB *IRDB,
   assert(IRDB != nullptr);
   // Assume, we already have validated the json schema
 
-  llvm::outs() << "Load precomputed points-to info from JSON\n";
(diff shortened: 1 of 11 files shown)


lib/PhasarLLVM/Pointer/LLVMAliasSet.cpp
```diff
@@ -100,7 +100,7 @@ LLVMAliasSet::LLVMAliasSet(LLVMProjectIRDB *IRDB,
   assert(IRDB != nullptr);
   // Assume, we already have validated the json schema
 
-  llvm::outs() << "Load precomputed points-to info from JSON\n";
+  // llvm::outs() << "Load precomputed points-to info from JSON\n";
```
MMory, line 103: maybe remove the line?

fabianbs96: done

MMory (approved):
lgtm, thanks!

PR #618, opened 2023-05-07
Enable Building Phasar with C++20

fabianbs96:
Enabling building phasar with C++20 is a low-hanging fruit...


MMory (approved):
lgtm, thanks!

PR #619, opened 2023-05-10
Explicit Call Graph

fabianbs96:
Currently, the call-graph is part of the ICFG implementation; this is fine as long as we only have one ICFG impl -- LLVMBasedICFG.

This PR createes an own class and interface for call-graphs that is being used within the ICFG.
This may be seen as preparation for different ICFG implementations that may share the call-graph data.

Should be merged after #617.


MMory (approved):
lgtm, thanks!

fabianbs96:
Added `blocked` label as #617 should be merged first

PR #620, opened 2023-05-15
Fix result[s]AtInLLVMSSA()

fabianbs96:
The functions `resultAtInLLVMSSA()` und `resultsAtInLLVMSSA()` in the `SolverResults` crash if called with a non-void instruction that happens to be a terminator of a BasicBlock (e.g. a `invoke` stmt).
As there seems to be no exact solution to computing the resultsAtInLLVMSSA of terminator instructions, this PR provides an approximation that works in most cases.


vulder (approved):
The design looks a bit over-complicated but if you want to have it as generic as possible :wink:

PR #621, opened 2023-05-21
Phasar DynLib and LTO

fabianbs96:
PhASAR 's codebase is growing continuously which makes it harder for externals to keep an overview, e.g. which libraries should be linked against, which headers to include (and where to find them).

This PR adds convenience headers that include *every* header from a particular subdirectory (plus one that includes *whole* phasar).
Additionally, phasar now builds a large library `libphasar.a` / `libphasar.so` by default that can be linked against.
The smaller "component-wise" libraries are still there for a more optimized build.


MMory (commented):
Would be good to have a hint somewhere that makes people include their newly added header(s) to the directory include headers. Coding guidelines?

MMory (approved):
lgtm, thanks!

MMory:
@vulder does this break vara? just wanna make sure it doesn't before we merge.

vulder (approved):
From a comparability standpoint, this works for us.
PS: I did not fully check the rest of the PR, just if it works in our in-tree setup. :+1:

PR #625, opened 2023-06-03
Misc Fixes II

fabianbs96:
Again some smaller changes that now include:

- Fix an endless recursion in `ConstantEdgeFunction::join`
- Add const functions in `psr::Table` to make SolverResults const-correct (not silently adding default constructed entries to the underlying table when just calling `resultAt`)
- Add ctor overload to `LLVMProjectIRDB` that takes a `llvm::MemoryBufferRef`. Useful if your application needs to do some computations on the *raw* IR data before parsing it (e.g. calculating some hash-code, etc.)
- Add a utility `hms` to support printing time-durations. Useful if you want to benchmark some analyses


MMory (approved):
thanks a lot, looks good!

PR #627, opened 2023-06-15
Bump dependencies

MMory:
bumping the submodules in external to the most recent versions.

@vulder could you please check whether this breaks vara before we merge? Thanks :)


vulder (commented):
Works for us

PR #630, opened 2023-06-15
Worklist for IDESolver

fabianbs96:
Now finally, we have it! A worklist for the `IDESolver`!

The IDESolver is no longer recursive, i.e., you no longer need to increase your stack-limit for analyzing anything non-trivial with IFDS/IDE.


MMory (approved):
lgtm, thanks! that was long overdue.

PR #632, opened 2023-06-17
Fix InitialSeeds for IFDS

### LINKED ISSUE #629: Possible error in function resultatinLLVMSSA()
- [√] I have searched open and closed issues for duplicates
  - [√] I made sure that I am not using an old project version (DO: pull Phasar, update git submodules, rebuild the project and check if the bug is still there)

----------------------------------------
The resultatinLLVMSSA() helps to deal with the problem that new generated fact will only vaild after the current instrcuction. For example, the load instruction, if %i is hold, the fact %0 will hold after the load instruction. The resultatinLLVMSSA() aims to solve this problem by querying the result at the next instruction.
`%0 = load i32, i32* %i, align 4; `
However, it seems not consider the problem that if the next insturction is a call statement. Due to the underlying theory (and respective implementation) not fact at call statement holds and will only holds at the corresponding return site, i.e. the statement after the call statement.

fabianbs96:
The default edge value of the `InitialSeeds` for IFDS analyses is `TOP` (the bottom value of the binary lattice) indicating "no information". This is incorrect as holding IFDS dataflow facts semantically have to be annotated with the top value of the binary lattice, which is called `BOTTOM` in phasar.
This issue prevents the `IDESolver` from computing IFDS results at certain call sites.

This PR solves that problem by correctly assigning the `BOTTOM` edge value to the initial seeds of IFDS analyses.

Should fix #629.

### DIFF
--- a/include/phasar/DataFlow/IfdsIde/InitialSeeds.h
+++ b/include/phasar/DataFlow/IfdsIde/InitialSeeds.h
@@ -32,7 +32,7 @@ public:
   InitialSeeds(const std::map<N, std::set<D>> &Seeds) {
     for (const auto &[Node, Facts] : Seeds) {
       for (const auto &Fact : Facts) {
-        this->Seeds[Node][Fact] = BinaryDomain::TOP;
+        this->Seeds[Node][Fact] = BinaryDomain::BOTTOM;
       }
     }
   }
@@ -42,7 +42,7 @@ public:
   template <typename LL = L,
             typename = std::enable_if_t<std::is_same_v<LL, BinaryDomain>>>
   void addSeed(N Node, D Fact) {
-    addSeed(Node, Fact, BinaryDomain::TOP);
+    addSeed(Node, Fact, BinaryDomain::BOTTOM);
   }
 
   void addSeed(N Node, D Fact, L Value) {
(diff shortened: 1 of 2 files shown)


Luweicai:
Hi, @fabianbs96. This patch fix the problem that preventing the IDESolver from computing IFDS results at certain call sites. 

But it seems casue a new problem: the new generated fact will have the value of `TOP` at no called position. 
```
%1 = add nsw i32 %0, 1;
%2 = add nsw i32 %1, 1;
call void @tt(i32 %3);
%5 = add nsw i32 %4, 1;
```
The result is:
```
N: %1 = add nsw i32 %0, 1;
-----------------------------------------------------
D: @zero_value | V: BOTTOM
D: i32 %0 | V: BOTTOM

N: %2 = add nsw i32 %1, 1;
-----------------------------------------------------
D: @zero_value | V: BOTTOM
D: i32 %0 | V: TOP
%1 = add nsw i32 %0, 1; | V: TOP

N: call void @tt(i32 %3);
-----------------------------------------------------
D: @zero_value | V: BOTTOM
D: i32 %0 | V: BOTTOM
%1 = add nsw i32 %0, 1; | V: BOTTOM

N: %5 = add nsw i32 %4, 1;
-----------------------------------------------------
D: @zero_value | V: BOTTOM
D: i32 %0 | V: TOP
D: %1 = add nsw i32 %0, 1; | V: TOP
D: %2 = add nsw i32 %1, 1; | V: TOP

MMory (approved):
lgtm, thanks!

PR #635, opened 2023-06-19
Fix TaintConfig Initial Seeds

fabianbs96:
The `makeInitialSeeds` function from `LLVMTaintConfig` may generate seeds at skipped (e.g., debug-) instructions leading the IDESolver to produce no IDE edge values generated from those seeds


MMory (approved):
lgtm, thanks!

PR #636, opened 2023-06-20
Bump json-schema-validator to fix build on ARM

fabianbs96:
The recent PR #627 broke the build on ARM due to a bug in the JSON Schema Validator. 

This PR bumps the schema validator to the commit that actually fixes the bug (unfortunately no tagged release yet).

@janniclas can you check whether this works on your system?


MMory (commented):
lgtm. Just would like @vulder to confirm it does not break vara.

vulder (approved):

PR #638, opened 2023-06-28
Lock-step Solving and Timeout for IDE Solver

fabianbs96:
Currently the `solve()` method on the `IDESolver` cannot be interrupted once started. This is not problem for analyzing small target programs, but is unacceptable for larger analyses that may run for days.

This PR adds fine-grained control over the solving process while still abstracting the solver internals.

The abstraction consists of three basic functions: `initialize`, `next` and `finalize`.
They are being called according to the below flow-chart:
```mermaid
flowchart LR
  initialize-->|true| next
  initialize-->|false| finalize
  next-->|true| next
  next -->|false|finalize
```

In addition to these basic APIs there are convenience functions like `solveUntil`, `solveWithTimeout`, etc. that build upon this model.
The original `solve` method has been rewritten and makes use of this model as well while keeping its original semantics.

Further, the solving methods (including `finalize`) now return a view into the solver results (or owning solver-results when called with an rvalue-reference) for allowing to be used in a more fluent way.


include/phasar/DataFlow/IfdsIde/Solver/IDESolverAPIMixin.h
```diff
@@ -0,0 +1,534 @@
...
+  /// Requires that initialize() has been called before once and returned true
+  /// and that all previous next() or nextN() calls returned true as well.
+  ///
+  /// \returns True, iff there are more steps to process before calling
+  /// finalize()
+  [[nodiscard]] bool nextN(size_t MaxNumIterations) {
+    PHASAR_LOG_LEVEL(DEBUG,
+                     "[nextN]: Next " << MaxNumIterations << " Iterations");
+
+    for (size_t I = 0, End = MaxNumIterations; I != End; ++I) {
```
MMory, line 58: Why introduce End here?

include/phasar/DataFlow/IfdsIde/Solver/IDESolverAPIMixin.h
```diff
@@ -0,0 +1,534 @@
...
+            std::chrono::steady_clock::time_point TimeStamp) {
+          return TimeStamp - Start >= Timeout;
+        };
+
+    return std::move(*this).continueUntil(CancellationRequested, Interval);
+  }
+
+  // -- Async cancellation
+
+  /// Solves the analysis problen and periodically checks whether
```
MMory, line 350: typo

include/phasar/DataFlow/IfdsIde/Solver/IDESolverAPIMixin.h
```diff
@@ -0,0 +1,534 @@
...
+          if constexpr (std::is_invocable_r_v<
+                            bool, CancellationRequest,
+                            std::chrono::steady_clock::time_point>) {
+            return std::invoke(CancellationRequested, TimeStamp);
+          } else {
+            return std::invoke(CancellationRequested);
+          }
+        };
+
+    size_t NumIterations = Interval.count() * 500;
```
MMory, line 461: where does this magic constant come from?

include/phasar/DataFlow/IfdsIde/Solver/IDESolverAPIMixin.h
```diff
@@ -0,0 +1,534 @@
...
+
+      // Adjust NumIterations
+      auto IterationsPerMilli = double(NumIterations) / DeltaTime.count();
+      auto NewNumIterations =
+          size_t(IterationsPerMilli * double(Interval.count()));
+      NumIterations = (NumIterations + 2 * NewNumIterations) / 3;
+    }
+    auto End = std::chrono::steady_clock::now();
+    return !IsCancellationRequested(End);
+  }
```
MMory, line 484: I wonder how robust this is. propagate() is not constant time, but not that diverse in its runtime. Still the question is what is the granularity by which we want to be able to stop. If this here is good enough, okay. But maybe the time checking could/should be done in a thread?

unittests/Utils/PAMMTest.cpp
```diff
@@ -29,9 +29,9 @@ class PAMMTest : public ::testing::Test {
 TEST_F(PAMMTest, HandleTimer) {
   //   PAMM &pamm = PAMM::getInstance();
   PAMM::getInstance().startTimer("timer1");
-  std::this_thread::sleep_for(std::chrono::milliseconds(1200));
+  std::this_thread::sleep_for(std::chrono::milliseconds(120));
   PAMM::getInstance().stopTimer("timer1");
-  EXPECT_GE(PAMM::getInstance().elapsedTime("timer1"), 1200U);
+  EXPECT_GE(PAMM::getInstance().elapsedTime("timer1"), 120U);
```
MMory, line 34: I am not a fan of sneaking this change into this PR.

unittests/Utils/PAMMTest.cpp
```diff
@@ -50,6 +50,8 @@ TEST_F(PAMMTest, HandleCounter) {
 }
 
 TEST_F(PAMMTest, HandleJSONOutput) {
+  GTEST_SKIP() << "We should really assert something in this test...";
+
```
MMory, line 54: same, but you are right.

MMory (approved):
lgtm, it is good to have this, thank you!

PR #640, opened 2023-07-04
Path Tracing

fabianbs96:


include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerBase.h
```diff
@@ -0,0 +1,189 @@
...
+    std::unique_ptr<char[]> Owner;
+    if (NumBytes > MaxNumBytesInStackBuf) {
+      Owner.reset((char *)Buf);
+    }
+
+    auto Cache = (vertex_t *)Buf;
+    std::uninitialized_fill_n(Cache, EquivSize, graph_traits_t::Invalid);
+
+    auto *WLConsumeBegin =
+        (std::pair<vertex_t, vertex_t> *)((vertex_t *)Buf + EquivSize);
```
MMory, line 128: this should be a reinterpret_cast and not a C-style cast

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerMixin.h
```diff
@@ -0,0 +1,365 @@
...
+
+        // llvm::errs() << "At Inst " << llvmIRToString(Inst) << "; Fact: " <<
+        // Fact
+        //              << '\n';
+        // llvm::errs();
+
+        {
+          std::error_code EC;
+          llvm::raw_fd_ostream ROS("explicitesg-err.dot", EC);
+          ESG.printAsDot(ROS);
```
MMory, line 91: Here we're dumping some stuff to a file with hardcoded name to the cwd. Shouldn't that rather be configurable or just printed to stderr?

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerMixin.h
```diff
@@ -0,0 +1,365 @@
...
+        }
+
+        llvm::errs()
+            << "> ESG written to "
+            << std::filesystem::canonical("explicitesg-err.dot").string()
+            << '\n';
+        llvm::errs().flush();
```
MMory, line 98: same

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerMixin.h
```diff
@@ -0,0 +1,365 @@
...
+    //     llvm::interleaveComma(PartialPath, ROS, [&ROS](const auto *Inst) {
+    //       ROS << psr::getMetaDataID(Inst);
+    //     });
+    //     ROS << " ]";
+    //     ROS.flush();
+    //     return Buf;
+    //   });
+
+    //   llvm::errs() << "Rev Paths DAG has " << Dag.Roots.size() << " roots\n";
+    // }
```
MMory, line 147: should we remove this commented code?

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerMixin.h
```diff
@@ -0,0 +1,365 @@
...
+      //     llvm::raw_string_ostream StrStr(Str);
+      //     StrStr << "[";
+      //     llvm::interleaveComma(Node, StrStr, [&StrStr](const auto *Inst) {
+      //       StrStr << getMetaDataID(Inst);
+      //     });
+      //     StrStr << "]";
+      //     return Str;
+      //   });
+      // }
+
```
MMory, line 173: same

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerMixin.h
```diff
@@ -0,0 +1,365 @@
...
+      auto Equiv = minimizeGraph(Dag);
+
+      // llvm::errs() << "Equiv:\n";
+      // llvm::errs() << "> Size: " << Equiv.getNumClasses() << '\n';
+      // for (size_t i = 0, end = Equiv.getNumClasses(); i < end; ++i) {
+      //   llvm::errs() << "[" << i << "] = " << Equiv[i] << '\n';
+      // }
```
MMory, line 180: same

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerMixin.h
```diff
@@ -0,0 +1,365 @@
...
+
+      Dag = createEquivalentGraphFrom(std::move(Dag), Equiv);
+      // Dag = Derived::reverseDAG(
+      //     std::move(Dag), [&Equiv](vertex_t Vtx) { return Equiv[Vtx]; },
+      //     Equiv.getNumClasses(), Config.DAGDepthThreshold);
+
```
MMory, line 186: same

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerMixin.h
```diff
@@ -0,0 +1,365 @@
...
+        //     llvm::raw_string_ostream StrStr(Str);
+        //     StrStr << "[";
+        //     llvm::interleaveComma(Node, StrStr, [&StrStr](const auto *Inst) {
+        //       StrStr << getMetaDataID(Inst);
+        //     });
+        //     StrStr << "]";
+        //     return Str;
+        //   });
+        // }
+
```
MMory, line 204: same

include/phasar/Utils/AdjacencyList.h
MMory, line 1: this is the new graph lib, isn't it? which version?

lib/PhasarLLVM/DataFlow/PathSensitivity/Z3BasedPathSensitivityManager.cpp
```diff
@@ -0,0 +1,595 @@
...
+        //              << "]: " << Ctx.NodeConstraints[Adj].to_string() <<
+        //              '\n';
+        // if (auto Constr =
+        //         LPC.getConstraintFromEdge(PartialPath.front(), AdjPP.back()))
+        //         {
+        //   llvm::errs() << ">> With EdgeConstraint: " << Constr->to_string()
+        //                << '\n';
+        // } else {
+        //   llvm::errs() << ">> Without EdgeConstraint\n";
+        // }
```
MMory, line 94: remove?

MMory (commented):
just some minor remarks

include/phasar/Utils/AdjacencyList.h
fabianbs96: This is part of the code the graphlib is inspired from. Once we have that ready, we can replace this code by it

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerMixin.h
fabianbs96: The problem here is that the explicit ESG frequently grows quite large and the commandline buffer might not be sufficient to show all output. So we write the ESG into a file instead. This happens in a fatal-error path only, so I don't think, it is a huge issue. 

Maybe we should write to a temp file instead? What do you think?

MMory: Yes, that might be better. Then we should also print the file name.

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerMixin.h
fabianbs96: Handled above

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerMixin.h
fabianbs96: done

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerBase.h
```diff
@@ -0,0 +1,190 @@
...
+    // - One worklist WLInsert of same type where we insert to
+
+    // We iterate over the graph in levels. Each level corresponds to the
+    // distance from a root node. We always have all nodes from a level inside a
+    // worklist
+    // -- The level we are processing is in WLConsume, the next level in
+    // WLInsert. This way, we can stop the process, when we have reached the
+    // MaxDepth
+
+    auto NumBytes =
```
MMory, line 109: in the discussion we believed that this never overflows since EquivSize is limited to the number of DAG nodes. Let's make sure this is correct. Maybe consider a check to be 100% sure.

MMory (approved):
lgtm, finally path tracing in PhASAR! Thanks for porting this over!

CMakeLists.txt
```diff
@@ -276,6 +280,21 @@ if(NOT LLVM_ENABLE_RTTI AND NOT PHASAR_IN_TREE)
   message(FATAL_ERROR "PhASAR requires a LLVM version that is built with RTTI")
 endif()
 
+# Z3 Solver
+if(PHASAR_USE_Z3)
+  # This z3-version is the same version LLVM requires; however, we cannot just use Z3 via the LLVM interface
+  # as it lacks some functionality (such as z3::expr::simplify()) that we require
+  find_package(Z3 4.7.1 REQUIRED)
```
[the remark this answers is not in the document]
fabianbs96: Have looked it up: [find_package](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature) allows to specify a minimum version or a version range. In case of a single version, you can provide the option `EXACT` to request an exact match instead.
So, this find_package invocation looks correct

include/phasar/DataFlow/PathSensitivity/PathSensitivityManagerBase.h
fabianbs96: Added an assert

PR #644, opened 2023-07-20
added descriptions to TaintConfigSchema.json to explain its semantics

StamesJames:
I started playing around with phasar a bit to look if I can use it on llvm code created by rust. At first I have used the analysis-config wrong so I wrote some descriptions into the schema. Maybe this can help future users too and I hope my descriptions are right.

### DIFF
--- a/config/TaintConfigSchema.json
+++ b/config/TaintConfigSchema.json
@@ -14,29 +14,36 @@ R"(
             "description": "Version of this taint configuration"
         },
         "functions": {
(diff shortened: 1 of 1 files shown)


MMory:
Hi @StamesJames , thank you for your contribution. I have committed a minor change, making all the first letters of the description text capitals. Do you agree with that? If so, I'll merge.

StamesJames:
ah ok thanks for fixing this. 
And I will write the tutorial and put in in an Issue :-)

PR #651, opened 2023-08-02
Fix global handling in type state analysis

MMory:
TODO: write a unit test that captures the fixed behaviour


lib/PhasarLLVM/DataFlow/IfdsIde/Problems/IDETypeStateAnalysis.cpp
```diff
@@ -207,6 +207,8 @@ auto IDETypeStateAnalysisBase::getCallToRetFlowFunction(
           return killManyFlows(getWMAliasesAndAllocas(Arg.get()));
         }
       }
+      return killFlowIf(
+          [](d_t Source) { return llvm::isa<llvm::Constant>(Source); });
```
fabianbs96, line 211: Is it sound to kill globals here eagerly? In case of multiple callees a later one in  `Callees` may be an unrelated declaration-only function. Then we cannot safely kill anything

MMory: you are right, we can only kill if none of the callees is declaration-only. Thanks for spotting this!

fabianbs96 (approved):
lgtm, thanks!

PR #652, opened 2023-08-02
Make NToString and friends free functions

fabianbs96:
Currently, the printers `NToString`, `DToString`, etc. are pure virtual functions  that have to be implemented by each analysis problem. However, the vast majority of all implementations of these functions is exactly the same (delegating to either `llvmIRToString()` or to `operator<<`) and independent from the analysis problem instance.

To simplify the usage of these functions as well as to simplify implementations of `IDETabulationProblem`, these printer functions are now *free functions* directly within the namespace `psr`. They now have a useful default implementation and can be easily extended via normal function overloading.

Note: The only painpoint currently is the Typestate Analysis -- the only analysis where `LToString` really depends on the analysis.
This PR, thus, slightly adapts the `l_t` type of the TSA to cope with this issue. A better solution is welcome and may come in its own PR.


include/phasar/DataFlow/Mono/Contexts/CallStringCTX.h
```diff
@@ -68,11 +68,10 @@ template <typename N, unsigned K> class CallStringCTX {
     return Lhs.cs < Rhs.cs;
   }
 
-  llvm::raw_ostream &print(llvm::raw_ostream &OS,
-                           const NodePrinterBase<N> &NP) const {
+  llvm::raw_ostream &print(llvm::raw_ostream &OS) const {
     OS << "Call string: [ ";
     for (auto C : CallString) {
-      NP.printNode(OS, C);
+      OS << NToString((C));
```
MMory, line 74: double parentheses?

include/phasar/Utils/TypeTraits.h
```diff
@@ -233,6 +251,12 @@ template <typename T> struct DefaultConstruct {
   }
 };
 
+template <typename T, typename = std::enable_if_t<has_adl_to_string_v<T>>>
+[[nodiscard]] decltype(auto) adl_to_string(const T &Val) {
+  using std::to_string;
```
MMory, line 256: why?

include/phasar/Utils/TypeTraits.h
```diff
@@ -66,6 +68,19 @@ template <typename T>
 struct has_str<T, decltype(std::declval<T>().str())> : std::true_type {
 }; // NOLINT
 
+template <typename T> struct has_adl_to_string {
+  template <typename TT = T, typename = decltype(std::string_view(
+                                 to_string(std::declval<TT>())))>
```
MMory, line 73: std::to_string?

MMory (approved):
looks good, thanks! Nice cleanup!

PR #653, opened 2023-08-02
Cleaning up IDETabulationProblem

fabianbs96:
This PR consists of the following cleanup tasks:
- `allTopFunction()` now has a proper default
- the flow-function templates (`identityFlow()`, `lambdaFlow()`, etc.) have been moved to their own class not requiring them to template on `d_t` and `container_type` anymore. This is one step towards easy switching between different `container_type`s in order to find out, which one performs best for a given data-flow problem


vulder (approved):
For my parts, this looks fine

PR #654, opened 2023-08-06
Template Typestate Analysis

fabianbs96:
The typestate analysis is the only analysis in phasar where one printer function (`LToString`) depends on the concrete analysis instance. This is, because the TypestateDescription defines, how a type state should be printed.
Moving to free-printers (#652) would then require to make the type state itself a virtual type that knows how it should printed. This is not only ugly, but may also have a performance impact.

To fix this issue, this PR makes the typestate analysis a template enabling type-safe type states.


MMory (approved):
lgtm, thanks!

PR #655, opened 2023-08-06
Fix and Update PAMM

fabianbs96:
In current PhASAR versions, the Performance Measurement Mechanism (PAMM) is broken -- it does not even compile.

This PR fixes PAMM such that it compiles, make the unittests assert something and adds slight improvements on PAMM's code.


MMory (approved):
lgtm, thank you!

PR #660, opened 2023-08-18
Fix Backward ICFG

fabianbs96:
Currently the backward ICFG cannot be used as it lacks the implementation of some required interface functions.
This PR adds this functionality by forwarding to LLVMBasedICFG.
In addition, we use explicit template instantiation to enable detecting such issues earlier.


MMory (approved):
lgtm, thanks!

PR #663, opened 2023-08-28
Don't link fat LLVM lib in myphasartool, needs to be transitively pulled in through phasar

MMory:


vulder (approved):
Should the linking be really required, we should put it behind a USE_FAT lib check

fabianbs96 (approved):
lgtm

PR #664, opened 2023-08-28
Fix handling of unbalanced returns for IIA

MMory:
For the change in the IDE solver I am not sure yet how this should be done properly, let's discuss that @fabianbs96


vulder (approved):
From my point of view, this should work. However, I cannot say that removing the isZeroValue check is save.

fabianbs96:
@MMory I suppose this is correct; otherwise we cannot propagate back seed parameters in an unbalanced return, however, I don't know the exact reason for this being introduced (this is already there in Heros). On the other hand, for what use case do you need this?

Regarding the null-check on `CallSite` in the ret-FF, I don't think, this is the way to go. We should rather change the one location in the IDESolver where the call-site is passed `nullptr` an extra API in the IDETabulationProblem that can be overridden if required (I think, this is required only very rarely). Further, as the IDESolver is generic over the to be analyzed IR, we should also be able to handle statement (`n_t`) types that are not nullable.

PR #665, opened 2023-09-04
Adds helper function to determine if a function was generated by phasar

vulder:
During global analysis, phasar generates helper function to correctly handle global ctors/dtors and other global code fragments. With the new checking function, users can determine whether a given function was generated by phasar or not, e.g., for global analysis.


MMory:
In the course of this we should use consts instead of magic string literals. Somewhere we use them, somewhere we don't. I'm gonna make a commit in this branch to fix this.

MMory:
@vulder @fabianbs96 could you have a look at the changes I just committed?

vulder:
Looks good overall, let's merge this

fabianbs96 (approved):
lgtm

PR #672, opened 2023-10-02
TaintConfig Serialization

fabianbs96:
Several headers of PhASAR #include nlohmann/json which requires consumers of phasar to integrate nlohmann/json into their codebases as well. This is not only inconvenient for users, it also ties us to nlohmann without an easy way of switching JSON libraries under the hood.

To fix this, we introduce an intermediate layer (for now just for the TaintConfig) that represents the same json data but hides the implementation details


MMory (approved):
lgtm, thanks!

Issue #552, opened 2022-12-12
On-The-Fly Reporting


fabianbs96:
## Story

In general, PhASAR supports two kinds of analyses:
- Generally useful analyses such as constant analysis and alias analysis (and in future points-to analysis)
- Analyses reporting undesired dataflows

While analyses of the first kind typically run as helper analysis for a different client analysis, the latter kind oftentimes runs in whole-program analysis (WPA) mode or module-wise analysis (MWA) mode and aims to find bugs of vulnerabilities in the software under analysis.

Currently, the only way in reporting analysis results is by either specializing the emitTextReport() function on the analysis problem or by relying on the raw SolverResults. While this is mostly sufficient for helper analyses, reporting analysis lack a common way of reporting results in a generic way.

So, propose the following: The analysis problem permits an event-based reporting solution, that (for example via callbacks) allows to report findings in a generic way. This also adds the benefit of on-the-fly reporting: With the emitTextReport and SolverResults strategy all analysis results are only available once the whole analysis run has completed. Intermediate results are not supported. However, in a bigger system with other components (or users) waiting for analysis results it may be beneficial to report (intermediate) results as they arrive.

### Prerequisites

* *issue x*

### Out Of Scope

- *A isn't part of this*

### Acceptance Criteria

The following acceptance criteria are defined regarding the result of this story:
*Clarify what the final result should be, not how to accomplish it.*
* [ ] The IFDSTabulationProblem and IDETabultation problem provide an opt-in solution for generic reporting of analysis findings
* [ ] The results reporting allows for reporting intermediate results

### Related Stories

* *none*

### Follow-up Stories

* *none*

PR #677, opened 2023-11-06
Analysis Printer

fabianbs96:
A step towards #552 + pretty printing analysis findings


include/phasar/PhasarLLVM/DataFlow/IfdsIde/Problems/IDETypeStateAnalysis.h
```diff
@@ -519,70 +519,41 @@ class IDETypeStateAnalysis
...
-                        OS << "Pred State: " << LToString(Res.second) << '\n';
-                      }
-                    }
-                  }
-                  OS << "============================\n";
-                } else {
-                  OS << "\nAlloca : " << DToString(Res.first)
-                     << "\nState  : " << LToString(Res.second) << '\n';
+                  Warning<IDETypeStateAnalysisDomain<TypeStateDescriptionTy>>
+                      War(&I, Res.first, TSD->error());
```
MMory, line 533: /War/Warn?

include/phasar/PhasarLLVM/Utils/AnalysisPrinterBase.h
```diff
@@ -0,0 +1,44 @@
...
+  d_t Fact;
+  l_t LatticeElement;
+
+  // Constructor
+  Warning(n_t Inst, d_t DfFact, l_t Lattice)
+      : Instr(std::move(Inst)), Fact(std::move(DfFact)),
+        LatticeElement(std::move(Lattice)) {}
+};
+
+template <typename AnalysisDomainTy> struct Results {
```
MMory, line 23: I wonder whether the name Results is a bit too generic. It describes results of a dataflow analysis, but we also have PT, TH, CG etc. which cannot be expressed with this one.

include/phasar/PhasarLLVM/Utils/AnalysisPrinterBase.h
```diff
@@ -0,0 +1,44 @@
+#ifndef PHASAR_PHASARLLVM_UTILS_ANALYSISPRINTERBASE_H
+#define PHASAR_PHASARLLVM_UTILS_ANALYSISPRINTERBASE_H
+
+#include "llvm/Support/raw_ostream.h"
+
+namespace psr {
+
+template <typename AnalysisDomainTy> struct Warning {
```
MMory, line 8: Would other/more categories make sense?

MMory (commented):
lgtm, just some minor stuff. Thank you!

include/phasar/PhasarLLVM/Utils/AnalysisPrinterBase.h
sritejakv: The intuition here is that, static analysis only reports warnings due to the sound approximations. For example, an Error would mean that it would definetly occur at runtime and we can only report an Error if we use precise information in our analyses. IMHO, I don't think, there is a need for other categories yet but would definitely be a requirement if we use runtime information.

MMory (approved):
lgtm, thank you! Another step towards more concise code in phasar!

PR #678, opened 2023-11-07
Fix unexpected subshell execution in bootstrap.sh

szsam:
For `(list)` in bash, `list` is executed in a subshell environment. Variable assignments do not remain in effect after the command completes.

Fix by using `{ list; }`, where `list` is executed in the current shell environment.


MMory (approved):
Thank you @szsam for your contribution, we appreciate it!

PR #683, opened 2023-11-16
Improve Logger Performance

fabianbs96:
The `IF_LOG_ENABLED` macro only checks for the logger being enabled, not for the enabled log level. This leads to drastical performance drops in the IDESolver when enabling logging -- even if just INFO log is enabled.

This PR fixes this issue by providing a new macro `IF_LOG_LEVEL_ENABLED` that additionally checks for the enabled log level.
Additionally, the logger is slightly refactored


MMory (approved):
lgtm, as discussed recently, thanks!

PR #673, opened 2023-10-02
Cleanup Build

fabianbs96:
PhASAR's CMakeLists are highly redundant making them sometimes hard to understand. 
This PR is a collection of minor changes that aim to improve the cmake build scripts and the CI as well as other minor build-related issues.

PR #684, opened 2023-11-20
Fix Build

fabianbs96:
Although #673 was a major step towards modernizing our cmake build, there are still issues left, some of them being addressed by this PR


MMory (approved):
looks good, thank you! Now the build works again in Gentoo with system-wide installed llvm/clang.

PR #687, opened 2023-12-05
Provide LLVM's CFL-based Alias Analyses

fabianbs96:
As LLVM has removed the CFL-based alias analyses (CFLAndersAliasAnalysis and CFLSteensAliasAnalysis) that PhASAR depends on , we cannot upgrade the LLVM version.
To enable upgrading, we *temporarly* provide a copy of these LLVM analyses in PhASAR's code base.


MMory (approved):
Thank you Fabian. I think also with respect to adhering to the license it should be fine.

PR #690, opened 2023-12-08
Allow LLVM beeing installed in a custom path

### LINKED ISSUE #689: CMake files do not respect the LLVM config files from the custom installation
- [X ] I have searched open and closed issues for duplicates
  - [ X] I made sure that I am not using an old project version (DO: pull PhASAR, update git submodules, rebuild the project and check if the bug is still there)

----------------------------------------

### Bug description
CMake  does not use the include directory of the self-built LLVM/clang.

### Steps to reproduce
  - Use an LLVM version greater than 14 due a [bug](https://github.com/llvm/llvm-project/issues/57224) when building it on Debian 12
  - Build steps
    - Run `bootstrap.sh` 
    - Manually
      - `CC=/fs/scratch/al/llvm-14/bin/clang CXX=/fs/scratch/al/llvm-14/bin/clang++ cmake -G Ninja -DCMAKE_BUILD_TYPE=Release  -DPHASAR_BUILD_UNITTESTS=true /fs/scratch/al/phasar/src`
      - `cmake --build . -j 1`

**Actual result:** Due to the newer LLVM, the build process aborts. This in turn shows the wrong include dir: `-I/usr/lib/llvm-14/include`.
```
/fs/scratch/al/llvm-14/bin/clang++ -DBOOST_ALL_NO_LIB -DBOOST_GRAPH_DYN_LINK -DBOOST_REGEX_DYN_LINK -DPHASAR_BUILD_DIR=\"/fs/scratch/al/phasar/src/build\" -DPHASAR_SRC_DIR=\"/fs/scratch/al/phasar/src\" -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -I/usr/lib/llvm-14/include -I/fs/scratch/al/phasar/src/include -I/fs/scratch/al/phasar/src/build/include -I/fs/scratch/al/phasar/src/external/json/include -I/fs/scratch/al/phasar/src/external/json-schema-validator/src -MP -fstack-protector-strong -ffunction-sections -fdata-sections -pipe -Wall -Wextra -Wno-unused-parameter -O3 -DNDEBUG -march=native -fPIC -fvisibility-inlines-hidden -std=c++17 -MD -MT lib/CMakeFiles/phasar.dir/LibPhasar.cpp.o -MF lib/CMakeFiles/phasar.dir/LibPhasar.cpp.o.d -o lib/CMakeFiles/phasar.dir/LibPhasar.cpp.o -c /fs/scratch/al/phasar/src/lib/LibPhasar.cpp
```
**Expected result:** The CMake files should honor the custom LLVM installation. The parameter `-I/usr/lib/llvm-14/include` should be replaced by `-I/fs/scratch/al/llvm-14//include`.

(**Important**  `/fs/scratch/al/llvm-14/` does contain a built of LLVM 17.0.6. The directory name is misleading.)

### Context (Environment)

  - **phasar:** \[b796f7d44681514f58a683a3a71ff17c94edb0c1]
  - **googletest:** \[b07f56ab7d87e36290001f6ed3d5d181fb83b6a5]
  - **json:** \[bc889afb4c5bf1c0d8ee29ef35eaaf4c8bef8a5d]
  - **json-schema-validator** \[491ac44026e08f31790f5cacffa62e168bb35e32]

**Operating System:**
  - [X ] Linux
  - [ ] Windows
  - [ ] macOS

**Build Type:**
  - [ X] cmake
  - [ X] bootstrap.sh
  - [ ] custom build

### Possible solution

The CMake variable `LLVM_INCLUDE_DIRS`, used in `cmake/phasar_macros.cmake:209` shall point to the aforementioned directory.
That variable is set in `/fs/scratch/al/llvm-14/lib/cmake/llvm/LLVMConfig.cmake`. I assume CMake must use the CMake files for LLVM from `/fs/scratch/al/llvm-14/lib/cmake`.

### Example files

**Files:**

flipreverse:
Currently, the custom build of LLVM must be located in /usr/local/llvm-14.
If not, CMake will either fail finding LLVMConfig.cmake or use a different LLVM installation, e.g., the system-wide one. This will, for example, result in an include path different from the custom LLVM installation.
This commit extends the CMake scripts to force the user to specify the variable PHASAR_LLVM_INSTALL_DIR that points to the LLVM installation directory.
This variable is also used to extend CMAKE_PREFIX_PATH which will make CMake use LLVMConfig.cmake from the custom build.
(Fixed #689)

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -5,8 +5,8 @@ set -eo pipefail
 source ./utils/safeCommandsSet.sh
 
 readonly PHASAR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-readonly PHASAR_INSTALL_DIR="/usr/local/phasar"
-readonly LLVM_INSTALL_DIR="/usr/local/llvm-14"
+PHASAR_INSTALL_DIR="/usr/local/phasar"
+LLVM_INSTALL_DIR="/usr/local/llvm-14"
 
 NUM_THREADS=$(nproc)
 LLVM_RELEASE=llvmorg-14.0.6
@@ -26,7 +26,8 @@ function usage {
     echo -e "\t-DBOOST_DIR=<path>\t\t- The directory where boost should be installed (optional)"
     echo -e "\t-DBOOST_VERSION=<string>\t- The desired boost version to install (optional)"
     echo -e "\t-DCMAKE_BUILD_TYPE=<string>\t- The build mode for building PhASAR. One of {Debug, RelWithDebInfo, Release} (default is Release)"
-    echo -e "\t-DPHASAR_INSTALL_DIR=<path>\t- The folder where to install PhASAR if --install is specified (default is /usr/local/phasar)"
+    echo -e "\t-DPHASAR_INSTALL_DIR=<path>\t- The folder where to install PhASAR if --install is specified (default is ${PHASAR_INSTALL_DIR})"
+    echo -e "\t-DLLVM_INSTALL_DIR=<path>\t- The folder where to install LLVM if --install is specified (default is ${LLVM_INSTALL_DIR})"
 }
 
 # Parsing command-line-parameters
@@ -80,6 +81,15 @@ case $key in
     ;;
     -DPHASAR_INSTALL_DIR=*)
     PHASAR_INSTALL_DIR="${key#*=}"
+    shift # past argument=value
+    ;;
+    -DLLVM_INSTALL_DIR)
+    LLVM_INSTALL_DIR="$2"
+    shift # past argument
+    shift # past value
+    ;;
+    -DLLVM_INSTALL_DIR=*)
+    LLVM_INSTALL_DIR="${key#*=}"
     shift # past argument=value
     ;;
     -h|--help)
@@ -157,6 +167,9 @@ tmp_dir=$(mktemp -d "llvm-build.XXXXXXXX" --tmpdir)
 rm -rf "${tmp_dir}"
 echo "dependencies successfully installed"
 
+# *Always* set the LLVM root to ensure the Phasar script uses the proper toolchain
+LLVM_PARAMS=-DLLVM_ROOT="${LLVM_INSTALL_DIR}"
+
 echo "Updating the submodules..."
 git submodule update --init
 echo "Submodules successfully updated"
@@ -169,7 +182,7 @@ export CXX=${LLVM_INSTALL_DIR}/bin/clang++
 
 mkdir -p "${PHASAR_DIR}"/build
 safe_cd "${PHASAR_DIR}"/build
-cmake -G Ninja -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" "${BOOST_PARAMS}" -DPHASAR_BUILD_UNITTESTS="${DO_UNIT_TEST}" "${PHASAR_DIR}"
+cmake -G Ninja -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" "${BOOST_PARAMS}" -DPHASAR_BUILD_UNITTESTS="${DO_UNIT_TEST}" "${LLVM_PARAMS}" "${PHASAR_DIR}"
 cmake --build . -j "${NUM_THREADS}"
 
 echo "phasar successfully built"


fabianbs96:
Hi @flipreverse, 

as I already commented on #689, PhASAR's CMake Configuration should be able to automatically find the correct LLVM installation.

In addition, the bootstrap script is meant for the very first build of phasar only and it installs all system dependencies of phasar, as well as the correct LLVM version.
If you use a custom LLVM that you have built (and installed) already, the bootstrap script is probably not, what you are looking for.

Furthermore, it is not a good idea to force users to pass an additional argument to the cmake command (PHASAR_LLVM_INSTALL_DIR).
So, as of now, I am *not* in favor of merging this PR.

flipreverse:
> Hi @flipreverse,
> 
> as I already commented on #689, PhASAR's CMake Configuration should be able to automatically find the correct LLVM installation.
> 
I can ensure you, it doesn't. It looks at system paths, e.g., `/usr/lib/...`,  and at `/usr/local/llvm-14/`.
There is currently **no** way to make CMake use a custom path. We, for example, want to install that LLVM custom build at `/fs/proj/phasar` . CMake won't detect that path unless explicitly specified.
> In addition, the bootstrap script is meant for the very first build of phasar only and it installs all system dependencies of phasar, as well as the correct LLVM version. If you use a custom LLVM that you have built (and installed) already, the bootstrap script is probably not, what you are looking for.
> 
> Furthermore, it is not a good idea to force users to pass an additional argument to the cmake command (PHASAR_LLVM_INSTALL_DIR). So, as of now, I am _not_ in favor of merging this PR.
What do you think about an optional argument?
`PHASAR_LLVM_INSTALL_DIR` defaults to `/usr/local/llvm-14`, and can be overwritten.

fabianbs96:
Hi @flipreverse,
according to the cmake  documentation for [find_package](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure) you can customize the search procedure. 
Have you tried setting the cmake- or env variable `LLVM_ROOT` to your LLVM installation?

flipreverse:
> Hi @flipreverse, according to the cmake documentation for [find_package](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure) you can customize the search procedure. Have you tried setting the cmake- or env variable `LLVM_ROOT` to your LLVM installation?

That did the trick. Thx! :)
Now I get a compile error on my Debian Testing: `use of undeclared identifier 'uint32_t'
  std::vector<uint32_t> SourceValues;`. But that is a different issue.

Out of curiosity: Why are `LLVM_INSTALL_DIR` and `PHASAR_INSTALL_DIR` read-only variables? At least for `PHASAR_INSTALL_DIR` there is a cmdline switch.

What do you think about the following patch?
Add a new cmdline switch to change the LLVM install dir. If used, bootstrap.sh will set `LLVM_ROOT`.

fabianbs96:
Hi @flipreverse, to answer your questions:

> Out of curiosity: Why are LLVM_INSTALL_DIR and PHASAR_INSTALL_DIR read-only variables? At least for PHASAR_INSTALL_DIR there is a cmdline switch.

We simply did not have the use-case yet of installing LLVM to a custom location; probably `PHASAR_INSTALL_DIR` should not be readonly then anyway

> What do you think about the following patch?
Add a new cmdline switch to change the LLVM install dir. If used, bootstrap.sh will set LLVM_ROOT

This sounds reasonable

flipreverse:
> Hi @flipreverse, to answer your questions:
> 
> > Out of curiosity: Why are LLVM_INSTALL_DIR and PHASAR_INSTALL_DIR read-only variables? At least for PHASAR_INSTALL_DIR there is a cmdline switch.
> 
> We simply did not have the use-case yet of installing LLVM to a custom location; probably `PHASAR_INSTALL_DIR` should not be readonly then anyway
> 
> > What do you think about the following patch?
> > Add a new cmdline switch to change the LLVM install dir. If used, bootstrap.sh will set LLVM_ROOT
> 
> This sounds reasonable

What is your opinion on the new commit?

bootstrap.sh
```diff
@@ -26,7 +26,8 @@ function usage {
     echo -e "\t-DBOOST_DIR=<path>\t\t- The directory where boost should be installed (optional)"
     echo -e "\t-DBOOST_VERSION=<string>\t- The desired boost version to install (optional)"
     echo -e "\t-DCMAKE_BUILD_TYPE=<string>\t- The build mode for building PhASAR. One of {Debug, RelWithDebInfo, Release} (default is Release)"
-    echo -e "\t-DPHASAR_INSTALL_DIR=<path>\t- The folder where to install PhASAR if --install is specified (default is /usr/local/phasar)"
+    echo -e "\t-DPHASAR_INSTALL_DIR=<path>\t- The folder where to install PhASAR if --install is specified (default is ${PHASAR_INSTALL_DIR})"
+    echo -e "\t-DLLVM_INSTALL_DIR=<path>\t- The folder where to install PhASAR if --install is specified (default is ${LLVM_INSTALL_DIR})"
```
fabianbs96, line 30:
```suggestion
    echo -e "\t-DLLVM_INSTALL_DIR=<path>\t- The folder where to install LLVM (default is ${LLVM_INSTALL_DIR})"
```

flipreverse: Oh, thx. Fixed.

fabianbs96 (approved):
lgtm, thanks!

MMory (approved):

PR #691, opened 2023-12-11
TaintConfigData need include cstdint for uint32_t

flipreverse:
On my system (Debian Testing aka 13/Trixie), I have to explicitly include cstdint to make Phasar compile.


fabianbs96 (approved):
lgtm, thanks!

PR #698, opened 2024-01-17
Misc Fixes

fabianbs96:
Fixes the initial values handling in IDESolver to handle the case that a seed location can be reached from another path.

Additionally fixes the MaybeUniquePtr with incomplete types


MMory (approved):
lgtm, thanks!

PR #699, opened 2024-02-08
Fix DebugLibDeps

fabianbs96:
The cmake option `PHASAR_DEBUG_LIBDEPS` does not correctly verify that all dependencies are resolved.

To fix this, we now use the linker flag `-Wl,-z,defs` that reports a linker error if a library contains unresolved symbols.
We also fix dependency errors that were newly discovered by this.


.github/workflows/ci.yml
```diff
@@ -16,14 +16,14 @@ jobs:
         build: [ Debug, Release ]
         include:
           - build: Debug
-            flags: -DPHASAR_BUILD_DYNLIB=ON -DPHASAR_ENABLE_SANITIZERS=ON
```
MMory, line 19: reverse the swap?

MMory (approved):
lgtm, good that you discovered and did this, thanks!

Issue #623, opened 2023-05-29
DI-based TypeHierarchy


fabianbs96:
## Story

The current `LLVMTypeHierarchy` implementation relies on unrealistic assumptions about the code structure in LLVM IR and is of only little use in practice. 
We need a more robust implementation that relies on debug-information.

### Prerequisites

* *issue x*

### Out Of Scope

- *A isn't part of this*

### Acceptance Criteria

The following acceptance criteria are defined regarding the result of this story:
*Clarify what the final result should be, not how to accomplish it.*
* [ ] DebugInfo-based TypeHierarchy implementation that works even for C++-based IR that was composed from different source-files via `llvm-link`
* [ ] *Criterion 2*

### Related Stories

* *none*

### Follow-up Stories

* *none*

PR #702, opened 2024-02-25
DI-based LLVM Type Hierarchy

fabianbs96:
#623


lib/PhasarLLVM/TypeHierarchy/DIBasedTypeHierarchy.cpp
```diff
@@ -0,0 +1,257 @@
...
+      } else {
+        Seen.set(Curr);
+      }
+      Hierarchy.push_back(VertexTypes[Curr]);
+      // llvm::errs() << " -- push " << VertexTypes[Curr]->getName() << '\n';
+      WorkList.push_back(~Curr);
+      WorkList.append(DerivedTypesOf[Curr].begin(), DerivedTypesOf[Curr].end());
+    }
+  }
+}
```
MMory, line 168: The constructor is rather large, probably it makes sense to decompose it a bit.

MMory (commented):
lgtm, except minor. thanks!

PR #706, opened 2024-03-04
Fix debug-libdeps in CI

fabianbs96:
The flag `PHASAR_DEBUG_LIBDEPS` ignored dependency issues arising from debug-only code, such as `assert`.


MMory (approved):
lgtm, thanks

Issue #566, opened 2023-01-15
phasar installs files from googletest


yurivict:
```
===>   Registering installation for phasar-1222
Installing phasar-1222...
pkg-static: phasar-1222 conflicts with googletest-1.12.1 (installs files into the same place).  Problematic file: /usr/local/include/gtest/gtest-death-test.h
*** Error code 1
```

Version: 1222

fabianbs96:
PhASAR installs these dependencies, because it needs them to compile and run. We are aware that version conflicts are a problem and we are currently completely overhauling our dependency management. See #516.

yurivict:
@fabianbs96 

Any chance to fix this?
All distros must have the same problem with conflict between nlohmann json versions.
This prevents porting efforts or makes them more difficult.

fabianbs96:
Hi @yurivict,
I understand that this is an issue, however, fixing this properly is not straightforward. Can you check whether #707 fixes your issue?

PR #707, opened 2024-03-04
Fix Dependendency Installation

fabianbs96:
Do not install phasar's dependencies to the global namespace anymore.

#566


MMory (approved):
lgtm, thanks!

PR #708, opened 2024-03-04
Improve phasar as subdirectory experience

fabianbs96:
PhASAR is recommended to be pulled in via `add_subdirectory` in cmake. However, there are many small things to improve in this scenario, mostly by excluding targets that are unneeded by clients. 
Additionally, phasar can then also better be used via FetchContent

We now also use FetchContent to pull in GoogleTest instead of including it as a submodule.
This avoids users to pull it in accidentally multiple times


MMory (approved):
looks good to me! I wasn't really aware of the FetchContent feature, cool stuff! At some point we might wanna bump the cmake version we require, we haven't done so in a while

PR #709, opened 2024-03-09
LLVM 15 Compatibility

fabianbs96:
Prepare for upgrading the LLVM version by fixing incompatibilities with LLVM 15.
Still, PhASAR works with LLVM 14 and the unittests still require typed pointers.


MMory (approved):
lgtm, thanks!

PR #714, opened 2024-04-07
Sparse IDE

fabianbs96:
Implementing Sparse IDE after Karakaya et al. https://doi.org/10.48550/arXiv.2401.14813

- [x] We NEED some handling of aliasing in the sparse CFG, otherwise the level of sparsity is too bad for analyzing LLVM IR


sritejakv (approved):
Looks good to me!

PR #717, opened 2024-04-20
Resolver Injection

fabianbs96:
Allows injecting an own `Resolver` implementation into the ctor of `LLVMBasedICFG`. ALso updates some other stuff in the resolvers


lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
```diff
@@ -40,11 +40,11 @@
 namespace psr {
 struct LLVMBasedICFG::Builder {
   LLVMProjectIRDB *IRDB = nullptr;
-  LLVMAliasInfoRef PT{};
+  // LLVMAliasInfoRef PT{};
```
MMory: just commented out?

MMory (commented):
looks good to me, just one minor remark

lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
fabianbs96: fixed

MMory (approved):
lgtm, thanks!

PR #723, opened 2024-05-10
Filtered AliasSet

fabianbs96:
The `LLVMAliasSet` contains a lot of spurious aliases due to its context insensitivity and its almost not present field-sensitivity.
Hence, the client analyses based on this alias information produce many false positives and need to propagate many more facts than necessary, resulting in a severe performance hit.

Therefore, some analyses in phasar, such as the `IFDSTaintAnalysis` and `IDEExtendedTaintAnalysis`, filter the alias sets on-the-fly to achieve better results.

This PR provides a wrapper `FilteredLLVMAliasSet` around the `LLVMAliasSet` that abstracts and caches the filtering.


include/phasar/PhasarLLVM/Pointer/FilteredLLVMAliasSet.h
```diff
@@ -0,0 +1,124 @@
...
+class LLVMAliasSet;
+class FilteredLLVMAliasSet;
+
+template <>
+struct AliasInfoTraits<FilteredLLVMAliasSet>
+    : DefaultAATraits<const llvm::Value *, const llvm::Instruction *> {};
+
+class FilteredLLVMAliasSet {
+public:
+  using traits_t = AliasInfoTraits<FilteredLLVMAliasSet>;
```
vulder: you might call it `alias_traits_t` so it easier to see what kind of traits one accesses in the following code.

include/phasar/PhasarLLVM/Pointer/FilteredLLVMAliasSet.h
```diff
@@ -0,0 +1,124 @@
...
+  using AliasSetTy = traits_t::AliasSetTy;
+  using AliasSetPtrTy = traits_t::AliasSetPtrTy;
+  using AllocationSiteSetPtrTy = traits_t::AllocationSiteSetPtrTy;
+
+  FilteredLLVMAliasSet(LLVMAliasSet *AS) noexcept;
+
+  FilteredLLVMAliasSet(const FilteredLLVMAliasSet &) = delete;
+  FilteredLLVMAliasSet &operator=(const FilteredLLVMAliasSet &) = delete;
+  FilteredLLVMAliasSet &operator=(FilteredLLVMAliasSet &&) = delete;
+
```
vulder, line 60:
```suggestion
```

include/phasar/PhasarLLVM/Pointer/FilteredLLVMAliasSet.h
```diff
@@ -0,0 +1,124 @@
...
+  using v_t = traits_t::v_t;
+  using AliasSetTy = traits_t::AliasSetTy;
+  using AliasSetPtrTy = traits_t::AliasSetPtrTy;
+  using AllocationSiteSetPtrTy = traits_t::AllocationSiteSetPtrTy;
+
+  FilteredLLVMAliasSet(LLVMAliasSet *AS) noexcept;
+
+  FilteredLLVMAliasSet(const FilteredLLVMAliasSet &) = delete;
+  FilteredLLVMAliasSet &operator=(const FilteredLLVMAliasSet &) = delete;
+  FilteredLLVMAliasSet &operator=(FilteredLLVMAliasSet &&) = delete;
```
vulder:
```suggestion
  FilteredLLVMAliasSet &operator=(FilteredLLVMAliasSet &&) noexcept = delete;
```

include/phasar/PhasarLLVM/Pointer/FilteredLLVMAliasSet.h
```diff
@@ -0,0 +1,124 @@
...
+  void printAsJson(llvm::raw_ostream &OS = llvm::outs()) const;
+
+  [[nodiscard]] AnalysisProperties getAnalysisProperties() const noexcept {
+    return AnalysisProperties::None;
+  }
+
+private:
+  FilteredLLVMAliasSet(MaybeUniquePtr<LLVMAliasSet, true> AS) noexcept;
+
+  MaybeUniquePtr<LLVMAliasSet, true> AS;
```
vulder:
```suggestion
  MaybeUniquePtr<LLVMAliasSet, true /*requiresAlignment*/> AS;
```

lib/PhasarLLVM/Pointer/FilteredLLVMAliasSet.cpp
```diff
@@ -0,0 +1,259 @@
...
+  if (const auto *Inst = llvm::dyn_cast<llvm::Instruction>(V)) {
+    return Inst->getFunction();
+  }
+  if (const auto *Arg = llvm::dyn_cast<llvm::Argument>(V)) {
+    return Arg->getParent();
+  }
+  return nullptr;
+}
+
+[[nodiscard]] static bool isConstantGlob(const llvm::GlobalValue *GlobV) {
```
vulder:
```suggestion
[[nodiscard]] static bool isConstantGlobalValue(const llvm::GlobalValue *GlobV) {
```
I find Glob not super easy to understand. I immediatly mixed it with shell globbing.

lib/PhasarLLVM/Pointer/FilteredLLVMAliasSet.cpp
```diff
@@ -0,0 +1,259 @@
...
+  if (const auto *Glob = llvm::dyn_cast<llvm::GlobalVariable>(GlobV)) {
+    return Glob->isConstant();
+  }
```
vulder, line 35: Why do you need a dyn_cast here? `GlobV` is already the type you want.

lib/PhasarLLVM/Pointer/FilteredLLVMAliasSet.cpp
```diff
@@ -0,0 +1,259 @@
...
+      continue;
+    }
+
+    if (V == Alias) {
+      Set.insert(Alias);
+      continue;
+    }
+
+    if (llvm::isa<llvm::ConstantExpr>(Alias) ||
+        llvm::isa<llvm::ConstantData>(Alias)) {
```
vulder:
```suggestion
    if (llvm::isa<llvm::ConstantExpr, llvm::ConstantData>(Alias)) {
```

vulder (changes requested):
Where do you implement the caching (I might have missed it :laughing: ) and how exactly do you ensure that the cache is a) not to big and b) how do you do cache invalidation.

include/phasar/PhasarLLVM/Pointer/FilteredLLVMAliasSet.h
fabianbs96: Does it make a difference to declare a deleted function `noexcept`?

include/phasar/PhasarLLVM/Pointer/FilteredLLVMAliasSet.h
fabianbs96: done

lib/PhasarLLVM/Pointer/FilteredLLVMAliasSet.cpp
fabianbs96: done

lib/PhasarLLVM/Pointer/FilteredLLVMAliasSet.cpp
fabianbs96: `GlobV` here is a `GlobalValue`, but I need it to be a `GlobalVariable` in order to call `isConstant()`.

lib/PhasarLLVM/Pointer/FilteredLLVMAliasSet.cpp
fabianbs96: Nice, thanks; I didn't know about that!

fabianbs96:
Hi @vulder, thanks for your comments.
To answer your questions:

> Where do you implement the caching (I might have missed it 😆 )

The caching is implemented in `getAliasSet()`, but now I have also implemented it for `getReachableAllocationSites()`

> and how exactly do you ensure that the cache is a) not to big and b) how do you do cache invalidation.

Currently, I have not implemented cache invalidation and just keep the cache growing. In my tests (on some coreutils), it has not been any problem. Also, for IFDS and IDE analyses, it is hard to predict, when an alias set is never used again. 

One could implement some ref-counting or least-recently-used strategy, but I havent' implemented it (yet?)

lib/PhasarLLVM/Pointer/FilteredLLVMAliasSet.cpp
vulder: Ah, sry I miss read that

include/phasar/PhasarLLVM/Pointer/FilteredLLVMAliasSet.h
vulder: Yes, a) you do the same thing for the move ctor and b) with noexcept is the correct signature of the function, so deleting it like this is the proper way

vulder:
> Hi @vulder, thanks for your comments. To answer your questions:
> 
> > Where do you implement the caching (I might have missed it 😆 )
> 
> The caching is implemented in `getAliasSet()`, but now I have also implemented it for `getReachableAllocationSites()`
> 
> > and how exactly do you ensure that the cache is a) not to big and b) how do you do cache invalidation.
> 
> Currently, I have not implemented cache invalidation and just keep the cache growing. In my tests (on some coreutils), it has not been any problem. Also, for IFDS and IDE analyses, it is hard to predict, when an alias set is never used again.
> 
> One could implement some ref-counting or least-recently-used strategy, but I havent' implemented it (yet?)

Ok, just keep in mind invalidation could also be interesting when the cache grows to big. So, even when a set would be needed again afterwards it would be useful to remove it from the cache to reduce the memory footprint.

lib/PhasarLLVM/DataFlow/IfdsIde/Problems/IFDSTaintAnalysis.cpp
```diff
@@ -155,7 +155,7 @@ void IFDSTaintAnalysis::populateWithMayAliases(
     container_type &Facts, const llvm::Instruction *Context) const {
```
MMory: The name Context is misleading/confusing me, context rather resembles a calling context, while here it is used as the statement at which the aliases are requested, in the sense of flow-sensitivity

MMory (approved):

sritejakv (approved):

PR #724, opened 2024-05-20
Move AnalysisController

fabianbs96:
The phasar-library `phasar_controller` is only used in `phasar-cli` and it is closely tied to it. Also, it does not fit well into our folder structure where all components that use `LLVMCore` should be in `PhasarLLVM`.
So, the controller better belongs into the application layer next to our tools.
As a bonus, clients that use phasar as a library no longer need to compile the controller if they do not need `phasar-cli` (which is very likely).


MMory (approved):
lgtm, thanks!

PR #725, opened 2024-05-20
Refactor getAsJson()

fabianbs96:
In the process to make `nlohmann::json` and implementation detail (and not having any nlohmann type in our public APIs), we:
1. Deprecate all `getAsJson()` functions
2. Add `printAsJson()` functions that should be used instead.


include/phasar/ControlFlow/CallGraphData.h
```diff
@@ -0,0 +1,33 @@
...
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace psr {
+struct CallGraphData {
+  std::unordered_map<std::string, std::vector<int>> FToFunctionVertexTy{};
```
MMory: are there assumption with respect to the format of the key, e.g. mangled(how?)/unmangled?

also, why vector<int>? Do we really want signedness here? The name suggests the value to be a vertex id, which I would expect to be unsigned. signed/unsigned mismatch is prone to causing unexpected trouble.

lib/PhasarLLVM/TypeHierarchy/DIBasedTypeHierarchy.cpp
```diff
@@ -167,11 +184,24 @@ DIBasedTypeHierarchy::DIBasedTypeHierarchy(const LLVMProjectIRDB &IRDB) {
 
     // -- Filter all struct- or class types
 
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
+    static constexpr llvm::dwarf::Tag DwarfTags[] = {
+        llvm::dwarf::DW_TAG_class_type,
+        llvm::dwarf::DW_TAG_structure_type,
+        llvm::dwarf::DW_TAG_union_type,
+    };
```
MMory, line 192: does it really have to be an array? what about `static constexpr auto DwarfTags = {`?

lib/PhasarLLVM/TypeHierarchy/DIBasedTypeHierarchy.cpp
```diff
@@ -189,6 +219,84 @@ DIBasedTypeHierarchy::DIBasedTypeHierarchy(const LLVMProjectIRDB &IRDB) {
...
+    if (Type->getName() == DITypeName) {
+      return Type;
+    }
+  }
+
+  llvm::report_fatal_error("DIType doesn't exist " + DITypeName);
+}
+
+// NOLINTNEXTLINE
+static constexpr char NullFunName[] = "__null__";
```
MMory: this is the second time I see this definition. Please use a single definition and make it static inline. also, again you can go for auto and avoid the c-style array.

MMory (approved):
please fix the minor things before merging.

include/phasar/ControlFlow/CallGraphData.h
fabianbs96: This is for unmangled names; I have clarified this as a comment.

Changed the signed int to unsigned; you are right, we don't need signedness here

lib/PhasarLLVM/TypeHierarchy/DIBasedTypeHierarchy.cpp
fabianbs96: If using `auto`, the type deduces to a `std::initializer_list` and I am unsure, whether it is safe to use this way (e.g., lifetimes, compiler-dependent bahavior, ...).

To be safe, I chose to use an array here.

lib/PhasarLLVM/TypeHierarchy/DIBasedTypeHierarchy.cpp
fabianbs96: fixed

PR #730, opened 2024-08-06
Opaque Pointers Integration

fabianbs96:
This PR adds the following changes:

- Integrate Opaque pointers into  Call-graph resolvers and unittests
- Disable the `DTAResolver` (to be removed for the next release).
- Replace the `LLVMTypeHierarchy` with `DIBasedTypeHierarchy` for all our analysis interfaces.
- Sets the default LLVM toolchain from 14 to 15
  - Also sets the CI to use LLVM 15


lib/PhasarLLVM/DataFlow/IfdsIde/Problems/IDEInstInteractionAnalysis.cpp
```diff
@@ -107,9 +109,7 @@ bool IDEIIAFlowFact::operator==(const IDEIIAFlowFact &Other) const {
   }
   auto EqualGEPDescriptor = [](const llvm::GetElementPtrInst *Lhs,
                                const llvm::GetElementPtrInst *Rhs) {
-    const auto *LhsI = llvm::dyn_cast<llvm::Instruction>(Lhs);
-    const auto *RhsI = llvm::dyn_cast<llvm::Instruction>(Rhs);
-    return LhsI->isSameOperationAs(RhsI);
+    return llvm::equal(Lhs->indices(), Rhs->indices());
```
MMory: also need to check scalar type here.

fabianbs96: done

sritejakv (approved):

Issue #720, opened 2024-04-29
Indices in uninit variables analysis


StamesJames:
I tried some phasar-cli analysis on rust code. On a simple hello-world program I got multiple unused variable notifications. When I looked at the llvm code those often where there because phasar hasn't tracked the indices that where already initialized and therefore defined an InsertValue operation as undefined as a whole, although just some indices are undefined.

I have started to implement a solution for this that I wanted to share here. I wrote a IndexWrapper class that acts as the dataflow fact and stores what indices inside a llvm::Value are the Facts.
While implementing this I also changed the current implementation of the uninit variables analysis to use the lambdaFlow function instead of local structs so maybe this could also be interesting for  PR #616

I'm currently not sure if I handle the GetElementPtrInst right and there is no alias analysis at the moment. 

I add my current implementations here:

[PhASAR-Uninit-Indexed-Variables.zip](https://github.com/secure-software-engineering/phasar/files/15148970/PhASAR-Uninit-Indexed-Variables.zip)

fabianbs96:
Hi @StamesJames, thank you for your contribution; it looks quite promising.

However, when reviewing the code, I realized, there are some issues:
- Memory Leaks: You allocate many objects with the `new` operator and not assigning an owner. This will likely exhaust your RAM quickly. Maybe consider using RAII types instead.
- Termination: In your flow functions you are allocating new data-flow facts (`IndexWrapper`) on the heap. Hence, when encountering a loop in the CFG, your program may constantly generate new facts (that all look the same) and not terminate. A better solution in this case may be using value types or using a deduplicating cache.
- Stack-use after Scope: in IndexWrapper.cpp:29 you create and store an ArrayRef initialized with a temporary reference to the parameter `index`. This will lead to a stack-use after scope error
- some other minor things

Can you maybe create a fork of phasar to put your implementation there? This would make reviews easier

StamesJames:
Thanks a lot for the review @fabianbs96. I'm quite new to C++ so I already thought there will be a lot of c++ beginner mistakes. I will make a fork and try to fix the issues.

StamesJames:
[...]

PR #732, opened 2024-09-24
Added a uninit variables analysis with indexes

StamesJames:
The added analysis exists as a seperate tool and does the same as the ifds-uninit analysis in phaser-cli but also remembers which indices are initialized.

This is related to Issue #720 

I started it because I tried using phaser for rust code and in rusts llvm code the compiler often initializes structs field by field because of which phaser detects a lot of uninitialized variables that are not really uninitialized.

### DIFF
--- a/include/phasar/PhasarLLVM/DataFlow/IfdsIde/Problems/IFDSUninitializedVariablesIndexed.h
+++ b/include/phasar/PhasarLLVM/DataFlow/IfdsIde/Problems/IFDSUninitializedVariablesIndexed.h
@@ -1,0 +1,102 @@
+/******************************************************************************
+ * Copyright (c) 2017 Philipp Schubert.
+ * All rights reserved. This program and the accompanying materials are made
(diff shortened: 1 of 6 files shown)


fabianbs96:
Hi @StamesJames,
thank you for your contribution! I am in favor of merging this into phasar.
For this to succeed, can you please modify your code to make the CI pipeline pass? The only thing it complained about was code formatting, so you can fix it by just running `clang-format-14` on the files that you have changed.

StamesJames:
Hi @fabianbs96 
yes sure I will fix this. And currently I haven't put the copyrights comment at the beginning of my newly created files and for the indexed analysis I just copied the original uninitialized analysis file so there is still the copyrights comment from the original. Should I change this somehow?

fabianbs96:
Hi @StamesJames,
yes, I think it is fine to keep the copyright notice. For the copied file, you can add yourself as contributor, if you like.
This should be fine from my point of view (I am not a lawyer, though^^)

PR #736, opened 2024-10-21
Make Compose and Join optional in EdgeFunction

fabianbs96:
Make the compose and join functions in the EdgeFunction implementations optional, if the IDETabulationProblem impl provides extend and combine.

Can we somehow emit a compile-time error, if neither are implemented?


include/phasar/DataFlow/IfdsIde/EdgeFunctionUtils.h
```diff
@@ -176,23 +179,24 @@ template <typename L> struct AllTop final {
...
+                                   const AllTop<L> &RHS) noexcept
     requires(!HasJoinLatticeTraits<L>)
   {
     return LHS.TopValue == RHS.TopValue;
   }
 };
 
 template <typename L, typename ConcreteEF>
-EdgeFunction<L>
+inline EdgeFunction<L>
```
vulder: Inline should not really be needed here

vulder (commented):
About your question reg. the compile error. Am I missing something, or would not just a OR + concept check in a static assert be enough? If you wanna talk about the design just ping me :)

Btw. I added Sebi to check if this conflicts with our internal setup.

vulder:
@boehmseb Can you please double check if this change would work for us.

fabianbs96:
The problem with concept+static_assert is that the Problem contains a virtual default implementation of extend and combine that forward to the EF-member-functions composeWith and joinWith. We would somehow need to statically detect that extend and combine have been overwritten.

I guess with the current design of IDETabulationProblem, this is not possible.

```C++
  virtual EdgeFunction<l_t> extend(const EdgeFunction<l_t> &L,
                                   const EdgeFunction<l_t> &R) {
    return L.composeWith(R); // <-- with the current design, this would emit a runtime error if compose() has
                             //     not been implemented *and* extend has not been overwritten
  }

  virtual EdgeFunction<l_t> combine(const EdgeFunction<l_t> &L,
                                    const EdgeFunction<l_t> &R) {
    return L.joinWith(R); // <-- similar possible runtime error as with compose
  }
```
Or do you have an idea, how one could detect that at compile time?

boehmseb:
> @boehmseb Can you please double check if this change would work for us.

Seems like it. At least our test still pass with this change.

PR #744, opened 2025-02-03
feat/conan2

jusito:
Added initial conan support with less changes compared to #516.

The Conanrecipe for llvm-core is mostly from conan center index but there is no 14.0.6 version, its planned to contribute to cci if I have the time. Best would be its part of cci, second best would be we would have an own public artifactory (currently running behind vpn) but having it here also works.

The easiest way of testing it is `utils/conan/test_conan_build.sh` because its testing the export itself and the usage of the recipe from cache, so we can be sure its working.

This PR allows to use an arbitrary compiler for build phasar with submodules and IR is generated with clang / clang++ / opt ver 14, if its available in PATH or in `/usr/local/llvm-14/bin`. However, this requires cmake >= 3.25.


cmake/phasar_macros.cmake
```diff
@@ -37,11 +37,94 @@ function(add_phasar_unittest test_name)
...
+function(validate_binary_version result item)
+  message(STATUS "  - validate_binary_version of \"${item}\"")
+  execute_process(
+    COMMAND "${item}" --version
+    OUTPUT_VARIABLE output
+    ERROR_VARIABLE error
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+
+  string(REGEX MATCH " version 14\\." match "${output}")
```
fabianbs96: Now that we have `PHASAR_LLVM_VERSION`, we can use it here to replace the 14

conanfile.py
```diff
@@ -0,0 +1,364 @@
...
+
+    return components
+
+class PhasarRecipe(ConanFile):
+    name = "phasar"
+    package_type = "library"
+
+    # Optional metadata
+    license = "MIT license"
+    author = "<Put your name here> <And your email here>"
```
fabianbs96: Are these TODOs?

test_package/example.ll
```diff
@@ -0,0 +1,170 @@
+; ModuleID = '/home/lucas/git/itst-develop/02_sa/intellisec-sast/analysis/test/resources/llvm_test_code/two_analyses_approach/ta_test_inter_true_positive.c'
+source_filename = "/home/lucas/git/itst-develop/02_sa/intellisec-sast/analysis/test/resources/llvm_test_code/two_analyses_approach/ta_test_inter_true_positive.c"
```
fabianbs96: Do you want to shorten these absolute paths?

conanfile.py
jusito: is optional and repository url is already provided, therefore removed https://docs.conan.io/2/reference/conanfile/attributes.html#author

test_package/example.ll
jusito: shortened

fabianbs96 (approved):

PR #749, opened 2025-02-17
ci: arm segfault

jusito:
The CI failed for arm64 build with segfault during compilation. We guessed that its an issue with outdated clang-14 and an clang upgrade would maybe fix it, the recent changed allowed this. [But it seems its a QEMU issue.](https://redirect.github.com/actions/partner-runner-images/issues/52)

The reworked Dockerfile is still a benefit but probably will not fix it. For that I increased the ubuntu version to 24.04, including the dependency installation script. Is there a specific distro / version we want to support - beside 20.04/24.04?

Some deps seems not directly used so removed?
- apt: graphviz
- pip3: Pygments pyyaml wllvm python3-sphinx


jusito:
@fabianbs96 
1. Changed os to 20.04 for the runner and created follow up issue #750 
2. Added PHASAR_TARGET_ARCH="" and removed PHASAR_BUILD_IR=$RUN_TESTS
3. It seems the expected jobs need to be updated.

fabianbs96 (approved):

PR #752, opened 2025-02-22
Update Phasar Config

fabianbs96:
Do not mention phasar's version at too many different places. Derive it from a single cmake variable.

Also, cleanup `PhasarConfig` a bit.


sritejakv (approved):
Looks good to me!

PR #753, opened 2025-02-22
Update Build

fabianbs96:
Smaller updates in the CMake configuration.

Update all built dependencies: nlohmann_json, json_schema_validator, and gtest.


sritejakv (approved):
Looks good to me!

PR #767, opened 2025-03-30
SVF Pointer Analyses

fabianbs96:
Integrate some pointer analyses from [SVF](https://github.com/SVF-tools/SVF) into PhASAR.

Currently, these comprise:
- VersionedFlowSensitive
- ContextDDA

We use type-erasure to make sure that no SVF types leak into PhASAR's headers.


include/phasar/Pointer/PointsToInfo.h
```diff
@@ -201,18 +191,12 @@ class PointsToInfoRef<PTATraits,
     return VT->GetPointsToSetV(PT, Pointer, AtInstruction);
   }
 
-  std::vector<v_t>
-  getInterestingPointersAtImpl(ByConstRef<n_t> AtInstruction) const {
```
sritejakv, line 205: There are two occurrences(DummyFieldInsensitivePointsToAnalysis, DummyFieldSensitivePointsToAnalysis) of this function definition which are unused. Should we remove them?

lib/PhasarLLVM/Pointer/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-file(GLOB_RECURSE POINTER_SRC *.h *.cpp)
+file(GLOB POINTER_SRC *.cpp TypeGraphs/*.cpp)
```
sritejakv: folder, "TypeGraphs" is not present in the repository

CMakeLists.txt
```diff
@@ -308,6 +308,18 @@ endif()
 include(add_llvm)
 add_llvm()
 
+# SVF
+option(PHASAR_USE_SVF "Use SVF for more options in alias analysis (default is OFF)" OFF)
```
sritejakv, line 312: We should also add some documentation about the steps involved in using SVF from PhASAR.

include/phasar/Pointer/PointsToInfo.h
fabianbs96: Yes, removed them

lib/PhasarLLVM/Pointer/CMakeLists.txt
fabianbs96: True, we recently removed it, thanks!

CMakeLists.txt
fabianbs96: Yes, we should add sth like that

sritejakv (approved):
Looks good to me!

PR #772, opened 2025-05-25
More Heap-Allocating Functions

fabianbs96:
Increase the set of know heap-allocating functions identified by `psr::isHeapAllocatingFunction()`


sritejakv (approved):
Looks good to me!

PR #773, opened 2025-05-25
Misc Changes

fabianbs96:
Adds a collection of smaller changes, including:

- Fix `AllBottom::compose`. For more information, see this related issue in Heros (https://github.com/soot-oss/heros/issues/37)
- Add API to `LLVMVFTableProvider` to retrieve the `llvm::GlobalVariable` for vtables
- Handle `atexit` in addition to `__cxa_atexit` in the `GlobalCtorsDtorsModel`
- Fix out-of-bounds access in `getVarTypeFromIR`
- more


lib/PhasarLLVM/DB/LLVMProjectIRDB.cpp
```diff
@@ -35,51 +38,128 @@ static void setOpaquePointersForCtx(llvm::LLVMContext &Ctx, bool Enable) {
...
+  }
+
+  [[nodiscard]] std::string message(int Value) const override {
+    switch (IRDBParsingError(Value)) {
+    case IRDBParsingError::CouldNotParse:
+      return "Could not parse LLVM IR";
+    case IRDBParsingError::CouldNotVerify:
+      return "Parsed LLVM IR could not be verified";
+    default:
+      "<invalid>";
```
sritejakv: Maybe update the message to "Unknown error while parsing IRDB" instead of "invalid".

lib/PhasarLLVM/DB/LLVMProjectIRDB.cpp
```diff
@@ -35,51 +38,128 @@ static void setOpaquePointersForCtx(llvm::LLVMContext &Ctx, bool Enable) {
...
+      llvm::MemoryBuffer::getFileOrSTDIN(IRFileName, /*IsText=*/true);
+  if (std::error_code EC = FileOrErr.getError()) {
+    return FileOrErr.getError();
+  }
+
+  return getParsedIRModuleOrErr(*FileOrErr.get(), Ctx);
+}
+
+std::unique_ptr<llvm::Module>
+LLVMProjectIRDB::getParsedIRModuleOrNull(llvm::MemoryBufferRef IRFileContent,
```
sritejakv, line 117: I like the 'getParsedIRModuleOrErr' more than 'getParsedIRModuleOrNull' due to the better readability and error messages.

Having said that, do we really need 'getParsedIRModuleOrNull' ? There are only four places where this function is used, twice in LLVMProjecTIRDB.cpp and twice in LLVMBasedICFGGlobCtorDtorTest.cpp. Can we replace them with 'getParsedIRModuleOrErr' and get rid of 'getParsedIRModuleOrNull' ?

lib/PhasarLLVM/Utils/LLVMIRToSrc.cpp
```diff
@@ -180,22 +181,37 @@ getOffsetAndBase(const llvm::Value *V) {
...
+    llvm::errs() << "<null>\n";
+  }
+
+  Ty->print(llvm::errs());
+}
+LLVM_DUMP_METHOD extern void dumpDIType(const llvm::DIDerivedType *Ty) {
+  dumpDIType(static_cast<const llvm::DIType *>(Ty));
+}
+
+} // namespace t2
```
sritejakv: I have two comments here. First, are we introducing a new namespace 'namespace t2' ? Second, these methods are not used anywhere in the repository. Is there any reason on why we added them here?

sritejakv (changes requested):

lib/PhasarLLVM/DB/LLVMProjectIRDB.cpp
fabianbs96: done

lib/PhasarLLVM/DB/LLVMProjectIRDB.cpp
fabianbs96: I have now replaced all uses of `getParsedIRModuleOrNull` with `getParsedIRModuleOrErr`, but I didn't remove the old function yet. It is marked as deprecated now, so we can safely remove it after the next phasar-release

lib/PhasarLLVM/Utils/LLVMIRToSrc.cpp
fabianbs96: The t2 is wrong, thanks for spotting that!

These functions are meant to be called from the debugger (that's the reason for `LLVM_DUMP_METHOD`), so I would like to keep them; however, I moved them now to `LLVMShorthands` where the other dump functions live.

sritejakv (approved):
Looks good to me! Thanks.

PR #774, opened 2025-05-25
Example Programs

fabianbs96:
Adds a set of example programs under `examples/how-to/` that show, how to use PhASAR.


sritejakv (approved):
Looks good to me! Thanks.

PR #776, opened 2025-06-07
Fix Type-based Call-Graphs with Multiple Inheritance

fabianbs96:
The RTAResolver and CHAResolver give wrong results in some case of multiple inheritance.

When virtually calling a function through a receiver of a non-primary base-class (that is not the first base in case of multiple inheritance) the analysis may falsely use the (correct) virtual index on the vtable of the *first base class*, even if the receiver is one of the secondary bases.

It may make sense to merge #773 first.


sritejakv (approved):
Looks good to me! Thanks.

PR #785, opened 2025-06-30
Call-Graph Improvements

fabianbs96:
A set of smaller changes related to call-graph resolvers


sritejakv (approved):
Looks good to me! Thanks.

PR #787, opened 2025-08-03
GCC Compatibility

fabianbs96:
Make phasar compile with gcc (g++-11 and g++-13) and uncover (and fix) two bugs related to json ser/de with LLVMAliasSet and LLVMBasedICFG that were not detected until now, because of nlohmann's implicit conversions feature.


CMakeLists.txt
```diff
@@ -82,6 +82,11 @@ option(PHASAR_BUILD_MODULES "Build C++20 modules for phasar" OFF)
 
 include(CheckCXXCompilerFlag)
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  # Match clang's behavior, when handling naming collisions, such as "using l_t = l_t;"
+  string(APPEND CMAKE_CXX_FLAGS " -fpermissive")
```
vulder: I would highly advice against adding permissive here. Permissive turns many actual errors into warnings/nothing. It is not recommended to be turned on.

Either we should fine another fix for the collisions or, even better, just fix them.

include/phasar/PhasarLLVM/DataFlow/IfdsIde/LLVMZeroValue.h
```diff
@@ -53,7 +57,7 @@ class LLVMZeroValue : public llvm::GlobalVariable {
 
   // NOLINTNEXTLINE(readability-identifier-naming)
   static constexpr auto isLLVMZeroValue = [](const llvm::Value *V) noexcept {
-    return V == getInstance();
+    return isZeroValueHelper(V);
```
vulder: What's the underlying reason to have a lambda here instead of a normal static member?

include/phasar/PhasarLLVM/DataFlow/IfdsIde/LLVMZeroValue.h
```diff
@@ -36,6 +36,10 @@ class LLVMZeroValue : public llvm::GlobalVariable {
   LLVMZeroValue(llvm::Module &Mod);
 
   static constexpr llvm::StringLiteral LLVMZeroValueInternalName = "zero_value";
+  static bool isZeroValueHelper(const llvm::Value *V) noexcept {
+    // Need this helper function to make gcc happy
```
vulder: What's the problem here for gcc?

include/phasar/PhasarLLVM/DataFlow/IfdsIde/Problems/IDEInstInteractionAnalysis.h
```diff
@@ -942,6 +942,14 @@ class IDEInstInteractionAnalysisT
 
   inline l_t join(l_t Lhs, l_t Rhs) override { return joinImpl(Lhs, Rhs); }
 
+  struct IIAAKillOrReplaceEF;
+  struct IIAAAddLabelsEF;
+  // These friend declarations are needed to make gcc happy
```
vulder: Could you maybe add the problem here in the comment, it's a bit hard to figure out why you needed this here.

include/phasar/PhasarLLVM/DataFlow/IfdsIde/Problems/IDEInstInteractionAnalysis.h
```diff
@@ -942,6 +942,14 @@ class IDEInstInteractionAnalysisT
...
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &OS,
```
vulder: I assume making them hidden friends was not an option because you would need the full def. for the implementation?

vulder (changes requested):

CMakeLists.txt
fabianbs96: Yes, I agree.
Fixed it

include/phasar/PhasarLLVM/DataFlow/IfdsIde/LLVMZeroValue.h
fabianbs96: The original reason was to make it easier to pass this as callback to other functions, such as in [here](https://github.com/secure-software-engineering/phasar/blob/development/lib/PhasarLLVM/DataFlow/IfdsIde/Problems/IDELinearConstantAnalysis.cpp#L308) or [here](https://github.com/secure-software-engineering/phasar/blob/development/examples/how-to/08-write-ide-analysis/simple.cpp#L83).

Not sure whether it is worth it in the end though.

include/phasar/PhasarLLVM/DataFlow/IfdsIde/LLVMZeroValue.h
fabianbs96: For gcc, the type LLVMZeroValue is incomplete in the lambda, so it cannot perform the implicit conversion to pointer-to-base for the comparison `V == getInstance()`

include/phasar/PhasarLLVM/DataFlow/IfdsIde/Problems/IDEInstInteractionAnalysis.h
fabianbs96: done

include/phasar/PhasarLLVM/DataFlow/IfdsIde/Problems/IDEInstInteractionAnalysis.h
fabianbs96: Actually, we *can* simplify this, as we only need the friend on the IIA analysis class, not on the edge-functions.
Thanks for pointing that out!

Changed it.

include/phasar/PhasarLLVM/DataFlow/IfdsIde/LLVMZeroValue.h
vulder: Hmm, not 100% sure here (github code review might hide things), but would this not just work in the same way with a static function?
```
struct Foo {
    static int foobar(int val) {
        return val * 2;
    }
};

template <typename Fn>
int call(Fn&& f) {
    return f(21);
}

int main() {
    return call(Foo::foobar);
}
```

If I missed something pls let me know. Then we could try to find a way to express this better.

fabianbs96: It would work, but then `foobar` would implicitly decay into a function-pointer, whereas a lambda has its own type.
This has two implications: (1) A function-pointer requires storage space when captured by a flow/edge function -- the captureless lambda plays well with `[[no_unique_address]]` and empty-base optimization. (2) The function-pointer is less likely to be inlined.

fabianbs96: I just noticed that we can simplify the lambda using our recently-added `fn` template: 

```diff
-  static constexpr auto isLLVMZeroValue = [](const llvm::Value *V) noexcept {
-    return isZeroValueHelper(V);
-  };
+  static constexpr auto isLLVMZeroValue = fn<isZeroValueHelper>;
```

vulder: I'm not sure if the lambda gets completely removed in this usage context but I also don't have strong opinion on the usage here, my primary goal was just to get rid of the gcc hackaround.

If the fn fixes it, fine for me.

vulder (approved):

PR #788, opened 2025-08-09
Remove boost from BitVectorSet

fabianbs96:
Now, all uses of boost are either replaced or deprecated; so, after PhASAR's next release, we can cet rid of boost as dependency


include/phasar/Utils/BitVectorSet.h
```diff
@@ -61,122 +64,68 @@ inline llvm::ArrayRef<uintptr_t> getWords(const llvm::SmallBitVector &BV,
...
-          break;
-        }
 
-        PosPtr = Position.right.find(NextIdx);
-
-        assert(PosPtr->first < Bits.size() &&
-               "pos_ptr->first index into BitVector out of range");
-      } while (!Bits[PosPtr->first]);
+    constexpr IteratorTy(typename BitVectorTy::const_set_bits_iterator It,
+                         fn_t<iterTransform> Txn = {}) noexcept
```
sritejakv, line 99: Do we need the second constructor parameter? As I understand, here we are initializing a static function, iterTransform. Can we do it directly at the variable declaration site: Line 122 ?

include/phasar/Utils/BitVectorSet.h
```diff
@@ -253,10 +194,9 @@ class BitVectorSet {
   }
 
   void erase(const T &Data) noexcept {
-    auto Search = Position.left.find(Data);
-    if (Search != Position.left.end()) {
-      if (Bits.size() > Search->second) {
-        Bits.reset(Search->second);
+    if (auto Idx = Position.getOrNull(Data)) {
+      if (Bits.size() > *Idx) {
```
sritejakv: Why are checking for an Index that is greater than the size of the Bitset ? Shouldn't it be the other way: check if the Index is less than Bits.size and reset it ?

include/phasar/Utils/BitVectorSet.h
```diff
@@ -281,12 +221,10 @@ class BitVectorSet {
   [[nodiscard]] bool find(const T &Data) const noexcept { return count(Data); }
 
   [[nodiscard]] size_t count(const T &Data) const noexcept {
-    auto Search = Position.left.find(Data);
-    if (Search != Position.left.end()) {
-      if (Bits.size() > Search->second) {
-        return Bits[Search->second];
-      }
+    if (auto Idx = Position.getOrNull(Data)) {
+      return (Bits.size() > *Idx && Bits.test(*Idx)) ? 1 : 0;
```
sritejakv: Same as the previous comment: check if the Index is less than Bits.size() and do Bits.test(Index) ?

sritejakv (changes requested):

include/phasar/Utils/BitVectorSet.h
fabianbs96: Initially, I wanted to use a `llvm::mapped_iterator`, but unfortunately, it does not work with the `set_bits_iterator` (See the out-commented type-alias above). For forward-compatibility with future versions that may be able to use `mapped_iterator` here, I introduced the second ctor parameter.
Currently, it is not needed and can be removed

include/phasar/Utils/BitVectorSet.h
fabianbs96: No, we are checking `Bits.size() > *Idx` <==> `*Idx < Bits.size()`, so whether the index is inbounds of the `Bits` bit-vector. If the index is out-of-bounds, we cannot call `reset` on it

include/phasar/Utils/BitVectorSet.h
fabianbs96: same here

include/phasar/Utils/BitVectorSet.h
sritejakv: Clear now! Thanks. Just for easy readability, should we consider swapping the condition to "*Idx < Bits.size()" ?

include/phasar/Utils/BitVectorSet.h
sritejakv: Okay. Let's then leave it as is to support forward-compatibility. Thanks.

include/phasar/Utils/BitVectorSet.h
fabianbs96: > should we consider swapping the condition to "*Idx < Bits.size()" ?

done

sritejakv (approved):
Looks good to me!

PR #790, opened 2025-09-01
Refactor AnalysisPrinter

fabianbs96:
Refactor AnalysisPrinter to take a `llvm::raw_ostream&` in `onFinalize()` and provide a sensible default for LLVM-based analyses.
With that, we can default `emitTextReport()` to just call `Printer->onFinalize(OS)`.

Fixes missing emit-text-report output in phasar-cli


sritejakv (approved):
Looks good to me! Thanks.

PR #825, opened 2026-04-11
Make most namespace-scope operators hidden friends

fabianbs96:


include/phasar/Utils/DOTGraph.h
```diff
@@ -103,11 +103,12 @@ struct DOTNode {
   DOTNode(std::string FName, std::string L, std::string SId, unsigned FId = 0,
           bool IsStmt = true, bool Isv = true);
   [[nodiscard]] std::string str(const std::string &Indent = "") const;
-};
 
-bool operator<(const DOTNode &Lhs, const DOTNode &Rhs);
-bool operator==(const DOTNode &Lhs, const DOTNode &Rhs);
-llvm::raw_ostream &operator<<(llvm::raw_ostream &OS, const DOTNode &Node);
+  friend bool operator<(const DOTNode &Lhs, const DOTNode &Rhs);
```
vulder, line 107: Note: without a direct implementation this is not a hidden friend 😉 
Please double check where it could make sense to move the impl. otherwise, keeping the free function is fine.

vulder (changes requested):

include/phasar/Utils/DOTGraph.h
fabianbs96: Oh, wasn't aware of that^^.
Thanks!

PR #849, opened 2026-07-06
Model Extern Callback Calls in LLVM Call Graphs

### LINKED ISSUE #786: Handling Extern Callbacks in CallGraph
PhASAR should be able to determine for known extern functions, that specific parameters are called.
This includes, but is not limited to:

- `pthread_create`
- `__kmpc_fork_teams `
- `__kmpc_fork_call`
- Any function declared with [callback](https://clang.llvm.org/docs/AttributeReference.html#callback)

We cannot just add artificial ICFG edges, as then the subsequent analyses don't know anymore, how to interpret the argument list of a affected call-site.
We should probably do some IR rewriting, similar as we already do for global ctors/dtors.

e345ee:
# Model Extern Callback Calls in LLVM Call Graphs

## Summary

Fixes #786.

This PR adds IR-level modeling for external callback brokers. The goal is to make callbacks passed to known external functions visible to PhASAR as regular LLVM call-sites, instead of encoding them as special call graph edges that later analyses would have to understand separately.

The first supported cases are `pthread_create`, `__kmpc_fork_call`, `__kmpc_fork_teams`, and functions annotated with LLVM/Clang `callback` metadata.

## Why

Some external functions receive a function pointer and call it internally. For example, `pthread_create` receives a worker function, but the actual call happens inside the pthread runtime, outside the analyzed LLVM IR.

Without modeling, PhASAR can see the call to `pthread_create`, but not a normal call-site to the worker function. This PR fixes that by generating a small model function for each supported callback broker call-site.

The generated model keeps the original external call and adds a regular call to the callback argument with the modeled argument list. This gives the call graph and later analyses ordinary LLVM IR to work with.

## Architecture

The solution follows the direction suggested in the issue: it uses IR rewriting rather than artificial ICFG edges.

For a broker call like `pthread_create(&thread, NULL, worker, arg)`, PhASAR generates a private model function and redirects the original call-site to it. Inside that model, PhASAR emits the original broker call, then emits a normal callback call, then returns the original broker result.

The rewrite is validated before the model is created. If the callback callee argument is missing, if a callback argument index is out of range, or if the callback arity does not match, the original call-site is left unchanged and a warning is emitted. This avoids creating misleading generated IR for malformed or unsupported call-sites.

## Files

The main changes are grouped as follows:

- `include/phasar/PhasarLLVM/ControlFlow/ExternCallbackModel.h`
- `lib/PhasarLLVM/ControlFlow/ExternCallbackModel.cpp`

  These files contain the new callback model. The header exposes the small public interface, while the implementation handles metadata parsing, known broker specifications, validation, model generation, and call-site rewriting.

  The model also preserves relevant properties of the original broker call when cloning it into the generated function, including calling convention, attributes, debug location, metadata, operand bundles, and tail-call kind.

- `include/phasar/PhasarLLVM/ControlFlow/LLVMBasedCallGraphBuilder.h`
- `lib/PhasarLLVM/ControlFlow/LLVMBasedCallGraphBuilder.cpp`

  These files connect the model to call graph construction. Mutable builder overloads may rewrite supported callback brokers before building the graph, while const overloads remain non-mutating.

  The documentation also notes that resolver/helper state should be created after rewriting if it needs to see generated model functions.

- `include/phasar/PhasarLLVM/ControlFlow/LLVMBasedICFG.h`
- `lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp`

  These files integrate callback rewriting into ICFG construction. Rewriting happens before PhASAR creates resolver/helper analyses, so generated model functions are already part of the IRDB. I removed the second rewrite from the shared initialization path because it only repeated the same idempotent full IR pass after the models had already been inserted.

- `test/llvm_test_code/call_graphs/*extern_callback*.c`
- `unittests/PhasarLLVM/ControlFlow/CMakeLists.txt`
- `unittests/PhasarLLVM/ControlFlow/LLVMBasedICFGExternCallbackTest.cpp`

  These files add regression coverage for the new feature. The tests cover direct and indirect callbacks, multiple callback broker call-sites, OpenMP brokers, metadata-based brokers, idempotent rewriting, mutable call graph builder integration, and invalid specs such as missing broker arguments, invalid callback argument indices, and mismatched callback arity.

- `lib/PhasarLLVM/Utils/LLVMShorthands.cpp`
- `unittests/PhasarLLVM/ControlFlow/LLVMBasedICFGGlobCtorDtorTest.cpp`

  These files stabilize an existing global constructor/destructor related test. The test no longer depends on hardcoded instruction ids, and the guard detection handles the guard-load pattern emitted by the local LLVM lowering. No platform-specific test skip is added.

## Tests

Ran the full unit test suite in Docker:

```text
ctest --test-dir cmake-build/docker-full --output-on-failure -j2
```

Result:

```text
100% tests passed, 0 tests failed out of 60
```

Also checked formatting and patch hygiene with `git diff --check` and `clang-format --dry-run --Werror`.

I would be happy to hear your feedback and address any requested changes.


e345ee:
Copilot left a few comments on the generated IR details, and I pushed a follow-up addressing them.
I changed the model so it no longer copies tail-call kind, uses `undef` for unknown modeled values, clears stale call-site attributes after rewriting, and makes the ctor/dtor test use the next non-debug instruction.
I re-ran the tests locally in Docker, and the full suite passes

include/phasar/PhasarLLVM/ControlFlow/LLVMBasedCallGraphBuilder.h
```diff
@@ -43,6 +46,22 @@ buildLLVMBasedCallGraph(const LLVMProjectIRDB &IRDB, Resolver &CGResolver,
 /// Constructs a call-graph using the given CGResolver to resolve indirect
 /// calls.
 ///
+/// May insert model functions for known external callback brokers before the
+/// call-graph is built.
+/// If CGResolver owns helper-analysis state that should include generated model
+/// functions, rewrite the IRDB before constructing that state or use a CGType
+/// overload.
+[[nodiscard]] LLVMBasedCallGraph
+buildLLVMBasedCallGraph(LLVMProjectIRDB &IRDB, Resolver &CGResolver,
```
fabianbs96: At a call-site, this overload is hardly distinguishable from the other overloads (taking the IRDB as const-ref) I would suggest using a different name to make explicit that this function has different behavior.

include/phasar/PhasarLLVM/ControlFlow/LLVMBasedCallGraphBuilder.h
```diff
@@ -59,6 +78,19 @@ buildLLVMBasedCallGraph(const LLVMProjectIRDB &IRDB, Resolver &CGResolver,
...
+/// Constructs a call-graph using the given CGResolver to resolve indirect
+/// calls.
+///
+/// May insert model functions for known external callback brokers before the
+/// call-graph is built.
+/// If CGResolver owns helper-analysis state that should include generated model
+/// functions, rewrite the IRDB before constructing that state or use a CGType
+/// overload.
+[[nodiscard]] LLVMBasedCallGraph
+buildLLVMBasedCallGraph(LLVMProjectIRDB &IRDB, Resolver &CGResolver,
```
fabianbs96: same

lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
```diff
@@ -39,9 +40,12 @@ void LLVMBasedICFG::initialize(LLVMProjectIRDB *IRDB, Resolver &CGResolver,
                                Soundness S, bool IncludeGlobals) {
   if (IncludeGlobals) {
     auto *EntryFun = GlobalCtorsDtorsModel::buildModel(*IRDB, EntryPoints);
-    this->CG = buildLLVMBasedCallGraph(*IRDB, CGResolver, {EntryFun}, S);
+    this->CG = buildLLVMBasedCallGraph(
```
fabianbs96: Rewriting is done before every call to initialize(). So, better call the version of `buildLLVMBasedCallGraph` that does this already.

unittests/PhasarLLVM/ControlFlow/CMakeLists.txt
```diff
@@ -18,3 +19,13 @@ set(LLVM_LINK_COMPONENTS Linker) # The CtorDtorTest needs the linker
 foreach(TEST_SRC ${ControlFlowSources})
   add_phasar_unittest(${TEST_SRC})
 endforeach(TEST_SRC)
+
+add_dependencies(LLVMBasedICFGExternCallbackTest
```
fabianbs96: Should not be necessary as all phasar unittests depend on `LLFileGeneration`

unittests/PhasarLLVM/ControlFlow/LLVMBasedICFGExternCallbackTest.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+namespace {
+
+const llvm::Function *getGeneratedModel(const LLVMProjectIRDB &IRDB,
+                                        llvm::StringRef BrokerName) {
+  std::string Prefix = (ExternCallbackModel::ModelPrefix + llvm::Twine('.') +
+                        BrokerName + llvm::Twine('.'))
+                           .str();
+  for (const auto *F : IRDB.getAllFunctions()) {
+    if (F->getName().startswith(Prefix)) {
```
fabianbs96: `.startswith()` is not compatible with more recent LLVM versions. Use `.starts_with()` instead.

unittests/PhasarLLVM/ControlFlow/LLVMBasedICFGExternCallbackTest.cpp
```diff
@@ -0,0 +1,331 @@
...
+}
+
+llvm::SmallVector<const llvm::Function *, 4>
+getGeneratedModels(const LLVMProjectIRDB &IRDB, llvm::StringRef BrokerName) {
+  llvm::SmallVector<const llvm::Function *, 4> Models;
+  std::string Prefix = (ExternCallbackModel::ModelPrefix + llvm::Twine('.') +
+                        BrokerName + llvm::Twine('.'))
+                           .str();
+  for (const auto *F : IRDB.getAllFunctions()) {
+    if (F->getName().startswith(Prefix)) {
```
fabianbs96: same

lib/PhasarLLVM/ControlFlow/ExternCallbackModel.cpp
```diff
@@ -0,0 +1,555 @@
...
+  return llvm::dyn_cast<llvm::Function>(V->stripPointerCastsAndAliases());
+}
+
+llvm::Type *getPointerType(llvm::LLVMContext &Ctx) {
+  return llvm::PointerType::get(Ctx, 0);
+}
+
+llvm::Type *getCallbackArgType(const llvm::CallBase &BrokerCall, int ArgNo) {
+  if (ArgNo < 0 || static_cast<unsigned>(ArgNo) >= BrokerCall.arg_size()) {
+    return getPointerType(BrokerCall.getContext());
```
fabianbs96, line 175: What is the reason for this fallback?

lib/PhasarLLVM/ControlFlow/ExternCallbackModel.cpp
```diff
@@ -0,0 +1,555 @@
...
+}
+
+llvm::Value *adaptValue(llvm::IRBuilder<> &IRB, llvm::Value *Value,
+                        llvm::Type *ExpectedTy) {
+  if (Value->getType() == ExpectedTy) {
+    return Value;
+  }
+
+  if (Value->getType()->isPointerTy() && ExpectedTy->isPointerTy()) {
+    return IRB.CreatePointerCast(Value, ExpectedTy);
```
fabianbs96: dead code?

unittests/PhasarLLVM/ControlFlow/LLVMBasedICFGGlobCtorDtorTest.cpp
```diff
@@ -335,10 +359,25 @@ TEST_F(LLVMBasedICFGGlobCtorDtorTest, LCATest4_1) {
...
-  const auto *LoadX = IRDB.getInstruction(20);
-  const auto *End = IRDB.getInstruction(22);
+  const auto *GetFoo = IRDB.getFunctionDefinition("_Z6getFoov");
+  ASSERT_NE(nullptr, GetFoo);
+  const auto *FooStorage = IRDB.getGlobalVariableDefinition("_ZZ6getFoovE3foo");
+  ASSERT_NE(nullptr, FooStorage);
+
+  const auto *LoadFoo = findLoadFrom(GetFoo, FooStorage);
+  ASSERT_NE(nullptr, LoadFoo);
+  const auto *FooGet = LoadFoo->getNextNonDebugInstruction();
```
fabianbs96: `getNextNonDebugInstruction()` is incompatible with newer LLVM versions. If you really need this behavior, wrap this into `#if LLVM_VERSION_MAJOR <= 18` and otherwise call `getNextNode()`.

unittests/PhasarLLVM/ControlFlow/LLVMBasedICFGExternCallbackTest.cpp
```diff
@@ -0,0 +1,331 @@
...
+  ASSERT_NE(nullptr, Callback);
+
+  const auto *CallbackCall = getCallbackCall(*Model, CG, *Callback);
+  ASSERT_NE(nullptr, CallbackCall);
+  EXPECT_TRUE(CallbackCall->isIndirectCall());
+}
+
+TEST(LLVMBasedICFGExternCallbackTest, InvalidMetadataCallbackArg) {
+  expectNoRewrite(R"(
+define void @sink(i32 %value) {
```
fabianbs96: Can you create c files for these IR snippets in test/llvm_test_code? Then the unit-test does not break if in the future the IR syntax changes.

fabianbs96 (changes requested):
Hi @e345ee, thanks for contributing this PR. Already looks quite good.
Only a few comments

lib/PhasarLLVM/ControlFlow/ExternCallbackModel.cpp
e345ee: It was meant to cover two cases: negative indexes are used by the internal OpenMP specs for synthetic gtid/bound pointer arguments, and the out-of-range check avoided indexing past the broker-call operands. Looking at it again, only the negative-index case should fall back to ptr; out-of-range positive indexes should be rejected instead. Tell me if I'm wrong

lib/PhasarLLVM/ControlFlow/ExternCallbackModel.cpp
e345ee: Yes sorry

lib/PhasarLLVM/ControlFlow/ExternCallbackModel.cpp
fabianbs96: Sounds good

lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
```diff
@@ -57,17 +55,14 @@ LLVMBasedICFG::LLVMBasedICFG(LLVMProjectIRDB *IRDB,
...
-    PTOwn = std::make_unique<LLVMAliasSet>(IRDB);
-    PT = PTOwn.asRef();
+  if (IncludeGlobals) {
+    auto *EntryFun = GlobalCtorsDtorsModel::buildModel(*IRDB, EntryPoints);
+    this->CG =
+        buildLLVMBasedCallGraph(*IRDB, CGType, {EntryFun}, TH, VTP, PT, S);
+  } else {
+    this->CG =
+        buildLLVMBasedCallGraph(*IRDB, CGType, EntryPoints, TH, VTP, PT, S);
   }
```
fabianbs96, line 65: Why is this ctor skipping `ExternCallbackModel::rewriteCalls(*IRDB);`?
Previous Resolver::create+initialize() seemed fine.

lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
e345ee: Done in f1fefb9. initialize() now calls buildLLVMBasedCallGraphWithExternCallbackModels(), so the callback rewrite is handled by the call graph builder instead of being done manually in initialize().

lib/PhasarLLVM/ControlFlow/LLVMBasedICFG.cpp
e345ee: This is still handled, but in the CallGraphAnalysisType overload now. That overload calls ExternCallbackModel::rewriteCalls(IRDB) before Resolver::create(...), so the generated model functions are already visible when the resolver/helper state is created.

fabianbs96: Oh okay, I missed that

fabianbs96 (approved):

